PR #177, opened 2016-04-19
Fix the bug of generating global dict when using dimension relation based on new load DML

Jay357089:
as issue #120 has shown, load table DML will be changed to INSERT INTO and ashok has raise a pr #169 to resolve this, based on thess changes, i design the dict generating from pre-defined dimensions as below:
1.the user should create a dimension table when he wanna use external-column file(no matter single or multiple columns) to generate dict. the DDL statement is just like #169 shows.
2.in the load part,  get the pre-defined dimensions data from dimension table and filter then generate global dict.
3. No using pre-defined dimension data in fact table to generate global dict.

since the most changes have been made in past pr, i make the further little changes based on pr #169 , pls check. @gvramana @Vimal-Das @ashokblend what is your idea?


QiangCai:
please do rebase

PR #191, opened 2016-04-20
[Issue-100]Drop Table DDL Support.

ravikiran23:
DDl for drop table has been supported confining to Hive Standards,

DDL is : 
DROP TABLE [IF EXISTS] table_name;


gvramana:
merged and reworked in #414

PR #231, opened 2016-04-23
Replaced carbonDef schema with CarbonTable for Aggregate table

ManoharVanam:
Replaced carbonDef schema with CarbonTable for Aggregate table


ManoharVanam:
These changes are not req

PR #299, opened 2016-05-03
fix get file path bug from local machine

eason-lyx:
1.file which has multiple blocks need to skip file path start with 'file:' when data load in local machine.
2.use data.preferredBufferSize


eason-lyx:
@jackylk this pr for big local file.if add testcase for this ,need to add a big data size csv file.
i have verify in local.

gvramana:
retest this please

eason-lyx:
@gvramana that is a bug in preferredBufferSize, this is no initialize by configuraion.
so currently,always 0. so using data.preferredBufferSize to fix this.
and for local file which has multiple block may has a big size csv file,do i need to add this to test case?

kumarvishal09:
retest this please

kumarvishal09:
retest this please

kumarvishal09:
LGTM

PR #340, opened 2016-05-07
[issue-339]balance blocks in data loading

eason-lyx:
the algorithm for blance block as following:
1.get the task host from dummy rdd
2.filter block location which no exist in task host,which support yarn mode
3.sort block list by location number,if the locations is empty, then append to end of list
4.assign  (totalBlockNum/nodeNum ) block to each node,only local block will assign,if local block is no enough,then mark how many no local block need to handle in this node.
5.after finished first round assignment, start assign one block by robin way .Only local block will assign,if the node no exist the local block,then mark it need to handle no local block times +1.
6.if (totalblocknum-assignBlockNum) <= sum(no local block time),start to assign no local block,until all block were assign.

in this algorithm ,consider data locality and block balance.

this pr is relative with #346 which raise by @ravikiran23 
i will add test case later.

please check @gvramana @ravikiran23

### DIFF
--- a/core/src/main/java/org/carbondata/core/load/BlockDetails.java
+++ b/core/src/main/java/org/carbondata/core/load/BlockDetails.java
@@ -26,12 +26,14 @@ import java.io.Serializable;
  */
 public class BlockDetails implements Serializable {
 
-  //block offset
+  // block offset
   private long blockOffset;
-  //block length
+  // block length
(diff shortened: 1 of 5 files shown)


eason-lyx:
retest this please

jackylk:
please rebase

PR #365, opened 2016-05-10
[Issue-359] Direct Dictionary Query Flow handling as per the latest implementation

mohammadshahidkhan:
1. Handled the direct dictionary support in detail query flow 
2. Corrected the UT failure
TODO To be  handled filter query and data retention


core/src/main/java/org/carbondata/query/carbon/executor/util/QueryUtil.java
```diff
@@ -278,19 +278,24 @@ public static void extractDimensionsAndMeasuresFromExpression(
...
+      if (CarbonUtil.hasEncoding(encodingList, Encoding.DICTIONARY) && !CarbonUtil
+          .hasEncoding(encodingList, Encoding.DIRECT_DICTIONARY)) {
         dictionaryDimensionFromQuery.add(dimAggInfo.get(i).getDim().getColumnId());
       }
     }
     for (int i = 0; i < customAggExpression.size(); i++) {
       List<CarbonColumn> referredColumns = customAggExpression.get(i).getReferredColumns();
       for (CarbonColumn column : referredColumns) {
-        if (CarbonUtil.hasEncoding(column.getEncoder(), Encoding.DICTIONARY)) {
+        if (CarbonUtil.hasEncoding(column.getEncoder(), Encoding.DICTIONARY) && !CarbonUtil
```
kumarvishal09: why we need to check both the conditions? When direct_dictionary is there no need to add dictionary in encoding for the column

mohammadshahidkhan: To differentiate between No Dictionary, Dictionary and Direct dictionary column, DICTIONARY must be checked together with the direct

mohammadshahidkhan:
Closing this issue as have raised new PR #458 in the public branch.

PR #371, opened 2016-05-10
[BUG] Distribute Query blocks to nodes, to balance load to tasks

gvramana:
Based on number of executor.cores per node, distribute blocks to each task. So that query load is evenly distributed to all tasks.


kumarvishal09:
retest this please

PR #377, opened 2016-05-11
[issue-376] Code clean up Task

ManoharVanam:
Removed TODO comments


core/src/main/java/org/carbondata/query/cache/QueryExecutorUtil.java
```diff
@@ -22,7 +22,6 @@
 import java.util.Arrays;
```
kumarvishal09: remove this class

ravipesala:
Please rebase it

jackylk:
Do these code going to be removed include the logic for aggregated table which we are going to add later?

ManoharVanam:
These changes are not required

PR #380, opened 2016-05-12
[BUG] Add exception handle in MDKeyGenStep

Jay357089:
the original code only handles CarbonDataWriterException, which causes other exceptions can not be throw and it is difficult for developer to find bug.
@gvramana @QiangCai   pls check

### DIFF
--- a/processing/src/main/java/org/carbondata/processing/mdkeygen/MDKeyGenStep.java
+++ b/processing/src/main/java/org/carbondata/processing/mdkeygen/MDKeyGenStep.java
@@ -188,13 +188,18 @@ public class MDKeyGenStep extends BaseStep {
         writeCounter++;
       }
     } catch (CarbonDataWriterException e) {
-      LOGGER.error(e, "Failed for: " + this.tableName);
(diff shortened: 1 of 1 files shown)


QiangCai:
LGTM.

processing/src/main/java/org/carbondata/processing/mdkeygen/MDKeyGenStep.java
```diff
@@ -188,13 +188,18 @@ public boolean processRow(StepMetaInterface smi, StepDataInterface sdi) throws K
         writeCounter++;
       }
     } catch (CarbonDataWriterException e) {
-      LOGGER.error(e, "Failed for: " + this.tableName);
+      LOGGER.error(e, "Failed for table: " + this.tableName);
```
QiangCai: @Jay357089
1 please add MDKeyGenStep to error message.
2  To different exception, error message should be different

Jay357089:
@QiangCai  close this pr. handle in pr 468

PR #387, opened 2016-05-12
[#387] Added range level index while reading the forward dict…

sujith71955:
While reading the members from forward cache range level scanning will be performed so that the cache reading will be faster while performing 'IN' query and join queries

### DIFF
--- a/core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryInfo.java
+++ b/core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryInfo.java
@@ -23,6 +23,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Stack;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/carbondata/core/cache/dictionary/ColumnDictionaryInfo.java
```diff
@@ -172,6 +174,43 @@ private int getSurrogateKeyFromDictionaryValue(byte[] key) {
...
+   * given value
+   *
+   * @param key to be searched
+   * @return
+   */
+  public int getIncrementalSurrogateKeyFromDictionary(byte[] key,
+      Stack<Integer> startIndexContainer) {
+    String filterKey = new String(key);
+    int low = startIndexContainer.pop();
+    List<Integer> sortedSurrogates = sortOrderReference.get();
```
gvramana: change this function to take list of sorted keys as input, take reference of sort index, search and give result at once

sujith71955:
Closed as handled part of #570

Issue #328, opened 2016-05-06
Data load failing when the measure data is null


foryou2030:
When i use carbon to load data, if the measure columns of data csv is "", such as:
          the cube schema is:
             DIMENSIONS (name string, age integer) MEASURES (grade numeric)
          and data csv is:
             jacky,36,

i got an exception:

```
 16/05/06 11:54:17 ERROR step.CarbonSliceMergerStep: [vardhan: Graph - Carbon Slice Mergervardhan][partitionID:vardhan] {0}
org.pentaho.di.core.exception.KettleException: 
org.carbondata.processing.merger.exeception.SliceMergerException: Problem while renaming inprogress folder to actual
Problem while renaming inprogress folder to actual

	at org.carbondata.processing.merger.step.CarbonSliceMergerStep.renameFolders(CarbonSliceMergerStep.java:170)
	at org.carbondata.processing.merger.step.CarbonSliceMergerStep.processRow(CarbonSliceMergerStep.java:119)
	at org.pentaho.di.trans.step.RunThread.run(RunThread.java:40)
	at java.lang.Thread.run(Thread.java:722)
Caused by: org.carbondata.processing.merger.exeception.SliceMergerException: Problem while renaming inprogress folder to actual
	at org.carbondata.processing.merger.step.CarbonSliceMergerStep.renameLoadFolderFromInProgressToNormal(CarbonSliceMergerStep.java:234)
	at org.carbondata.processing.merger.step.CarbonSliceMergerStep.renameFolders(CarbonSliceMergerStep.java:163)
	... 3 more
16/05/06 11:54:17 INFO csvload.DataGraphExecuter: [Executor task launch worker-1][partitionID:default_vardhan_d538894c-7794-4ae7-8e6a-4dc4cb50772c] Graph execution is finished.
16/05/06 11:54:17 ERROR csvload.DataGraphExecuter: [Executor task launch worker-1][partitionID:default_vardhan_d538894c-7794-4ae7-8e6a-4dc4cb50772c] Graph Execution had errors
16/05/06 11:54:17 ERROR rdd.CarbonDataLoadRDD: [Executor task launch worker-1][partitionID:default_vardhan_d538894c-7794-4ae7-8e6a-4dc4cb50772c] {0}
org.carbondata.processing.etl.DataLoadingException: Internal Errors
	at org.carbondata.processing.csvload.DataGraphExecuter.execute(DataGraphExecuter.java:263)
[...]

PR #394, opened 2016-05-13
[Issue-328] Data load failing when the measure data is null

foryou2030:
fixed issue #328 
1 change calculateMaxMinUnique method name to calculateMaxMin 
2 null value don't need to call calculateMaxMin. fixed the load exception.

what's more, null measure value was stored as 0 , maybe null value should be stored as unique value.

@Vimal-Das @gvramana ,pls check it. thanks.


jackylk:
Can you add a testcase to verify it

QiangCai:
@foryou2030 
please check with @kumarvishal09

PR #396, opened 2016-05-13
auto identify high cardinality column in first load

QiangCai:
1  identify high cardinality column in first load
   if a column satisfy following formulas, this column is high cardinality column.
   column cardinality /  row count  >  70% (configurable)
   column cardinality / 1 million(configurable) > 70% (configurable)

2  don't generate dictionary for high cardinality column 

3  update column metadata  to no_dictionary


kumarvishal09:
In incremental load column which is having less data can increase by substantial incremental data load  then this prediction will go wrong better let user define  no dictionary while creating the table, i think no need to change current design

sraghunandan:
I agree @kumarvishal. @jackyli,@gvramana,@vimal-Das pls comment

jarray888:
if user don't specify no dictionary to a high carnality column , according our design , we will put all the dictionary to memory and cause our system hang or cause memory issues . 

we can refer some DB design , not need user to input which column need to do dictionary or not , the system can automatic decided .

ravipesala:
@jarray888 Yes, As per our current design user needs to input high cardinality column. This PR solves only some part of our problem because most of the cases first load may not able to find out the high cardinality columns unless it is big load. 
We may need to calculate high cardinality columns from multiple loads and mark them as high cardinality to find out properly.So it would become restructure call and query flow also needs to handle it.
And what if user intentionally wants high cardinality columns also as dictionary column for performance reasons?.

ravipesala:
@QiangCai Please rebase it.

vishalmobikwik:
+ Real @kumarvishal09

kumarvishal:
+ Adding Real @kumarvishal09

jarray888:
@ravipesala this is a temp solution only , to avoid system hang before we push a final solution .  i thought we can do some locality dictionary for high cardinality  like parquet  , or use hybrid dictionary , some do dictionary ,exceed some  threshold we don't do dictionary .

jarray888:
if we do a final solution ,  it is a little big change for carbon , need time to do it , before that , we should have some alternative solution to avoid system hang .

QiangCai:
pls review

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
```diff
@@ -337,6 +367,34 @@ class CarbonMetastoreCatalog(hive: HiveContext, val storePath: String, client: C
     schemaMetadataPath
   }
 
+  def updateMetadataByWrapperTable(
```
ravipesala: Make it private function

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
```diff
@@ -337,6 +367,34 @@ class CarbonMetastoreCatalog(hive: HiveContext, val storePath: String, client: C
...
+
+  def updateMetadataByThriftTable(schemaFilePath: String,
+      tableInfo: TableInfo, dbName: String, tableName: String): Unit = {
+
+    tableInfo.getFact_table.getSchema_evolution.getSchema_evolution_history.get(0)
+      .setTime_stamp(System.currentTimeMillis())
+    val schemaConverter = new ThriftWrapperSchemaConverterImpl
+    val wrapperTableInfo = schemaConverter
+      .fromExternalToWrapperTableInfo(tableInfo, dbName, tableName)
+    val carbonTableIdentifier = new CarbonTableIdentifier(dbName, tableName)
```
ravipesala: Remove unused variable `carbonTableIdentifier`

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -385,14 +432,68 @@ object GlobalDictionaryUtil extends Logging {
...
+    // write TableInfo
+    CarbonMetastoreCatalog.writeThriftTableToSchemaFile(schemaFilePath, tableInfo)
+
+    // update Metadata
+    val catalog = CarbonEnv.getInstance(sqlContext).carbonCatalog
+    catalog.updateMetadataByThriftTable(schemaFilePath, tableInfo,
+        model.table.getDatabaseName, model.table.getTableName)
+
+    // update CarbonDataLoadSchema
+    val carbonTabel = catalog.lookupRelation1(Option(model.table.getDatabaseName),
```
ravipesala: NIT Pls correct typo `carbonTabel`

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -385,14 +432,68 @@ object GlobalDictionaryUtil extends Logging {
...
    */
-  private def checkStatus(status: Array[(String, String)]) = {
-    if (status.exists(x => CarbonCommonConstants.STORE_LOADSTATUS_FAILURE.equals(x._2))) {
+  private def checkStatus(carbonLoadModel: CarbonLoadModel,
+                          sqlContext: SQLContext,
+                          model: DictionaryLoadModel,
+                          status: Array[(Int, String, Boolean)]) = {
+    var result = false
+    val noDictionaryColumns = new ArrayBuffer[CarbonDimension]
+    val databaseName = model.table.getDatabaseName
```
ravipesala: This variable is never used

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -385,14 +432,68 @@ object GlobalDictionaryUtil extends Logging {
...
+    val tableName = model.table.getTableName
+    status.foreach { x =>
+      val columnName = model.primDimensions(x._1).getColName
+      if (CarbonCommonConstants.STORE_LOADSTATUS_FAILURE.equals(x._2)) {
+        result = true
+        logError(s"table:$tableName column:$columnName generate global dictionary file failed")
+      }
+      if (x._3) noDictionaryColumns +=  model.primDimensions(x._1)
+    }
+    if (noDictionaryColumns.size >= 1) {
```
ravipesala: use `nonEmpty`

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -289,15 +314,37 @@ object GlobalDictionaryUtil extends Logging {
...
+      CarbonDataRDDFactory.readLoadMetadataDetails(carbonLoadModel, hdfsLocation)
+    }
     new DictionaryLoadModel(table,
       dimensions,
       hdfsLocation,
       dictfolderPath,
       dictFilePaths,
       dictFileExists,
-      dimensions.map(_.isComplex == true),
+      isComplexes.map { x => x }.toArray,
```
ravipesala: Directly use `isComplexes.toArray`

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -289,15 +314,37 @@ object GlobalDictionaryUtil extends Logging {
       dictFilePaths(f._2) = carbonTablePath.getDictionaryFilePath(f._1.getColumnId)
       dictFileExists(f._2) = CarbonUtil.isFileExists(dictFilePaths(f._2))
     }
+
+    // load high cardinality identify configure
```
ravipesala: Please keep them in try catch so if any datatype conversion fails then use default instead of load failure.

ravipesala: Or else better add this inside `CarbonProperties` to validate each property.

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -259,6 +271,17 @@ object GlobalDictionaryUtil extends Logging {
     }
   }
 
+  def isHighCardinalityColumn(columnCardinality: Int,
+                              rowCount: Long,
+                              model: DictionaryLoadModel): Boolean = {
+    if (columnCardinality > 0 && rowCount > 0 && model.highCardThreshold > 0) {
+      (columnCardinality.toDouble / rowCount * 100 > model.rowCountPercentage) &&
+        (columnCardinality.toDouble / model.highCardThreshold * 100 > model.thresholdPercentage)
```
ravipesala: Using so many parameters to find high cardinality column is very confusing for user to configure.
Can you simplify it by using any one of the logic like either 
'columnCardinality.toDouble / rowCount * 100 > model.rowCountPercentage'
or
`columnCardinality.toDouble / model.highCardThreshold * 100 > model.thresholdPercentage`

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -128,14 +129,21 @@ case class DataFormat(delimiters: Array[String],
...
+                               dictFileExists: Array[Boolean],
+                               isComplexes: Array[Boolean],
+                               primDimensions: Array[CarbonDimension],
+                               delimiters: Array[String],
+                               highCardIdentifyEnable: Boolean,
+                               highCardThreshold: Int,
+                               rowCountPercentage: Double,
+                               thresholdPercentage: Double,
+                               isFirstLoad: Boolean) extends Serializable
+
```
ravipesala: Please correct the method parameter indentation.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -147,15 +155,16 @@ case class DictionaryLoadModel(table: CarbonTableIdentifier,
...
-  extends RDD[(Int, Array[String])](prev) with Logging {
+  extends RDD[(Int, ColumnDistinctValues)](prev) with Logging {
 
   override def getPartitions: Array[Partition] = firstParent[Row].partitions
 
   override def compute(split: Partition, context: TaskContext
-  ): Iterator[(Int, Array[String])] = {
-    val LOGGER = LogServiceFactory.getLogService(this.getClass.getName)
+    ): Iterator[(Int, ColumnDistinctValues)] = {
+    val LOGGER = LogServiceFactory.getLogService(this.getClass().getName)
```
ravipesala: Please use like below for correct style
override def compute(split: Partition, 
context: TaskContext): Iterator[(Int, ColumnDistinctValues)] = {

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -147,15 +155,16 @@ case class DictionaryLoadModel(table: CarbonTableIdentifier,
...
     model: DictionaryLoadModel)
-  extends RDD[(Int, Array[String])](prev) with Logging {
+  extends RDD[(Int, ColumnDistinctValues)](prev) with Logging {
 
   override def getPartitions: Array[Partition] = firstParent[Row].partitions
 
   override def compute(split: Partition, context: TaskContext
-  ): Iterator[(Int, Array[String])] = {
-    val LOGGER = LogServiceFactory.getLogService(this.getClass.getName)
+    ): Iterator[(Int, ColumnDistinctValues)] = {
```
ravipesala: Please use like below for correct style
override def compute(split: Partition, 
context: TaskContext): Iterator[(Int, ColumnDistinctValues)] = {

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
ravipesala: remove parenthesis for `this.getClass()`. make like `this.getClass`

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -228,36 +239,53 @@ class CarbonGlobalDictionaryGenerateRDD(
...
+        breakable {
+          while (rddIter.hasNext) {
+            val distinctValueList = rddIter.next()._2
+            valuesBuffer ++= distinctValueList.values
+
+            // check high cardinality
+            if (model.isFirstLoad && model.highCardIdentifyEnable &&
+                !model.isComplexes(split.index)) {
+              rowCount += distinctValueList.rowCount
+              isHighCardinalityColumn = GlobalDictionaryUtil.isHighCardinalityColumn(
```
ravipesala: Better calculate 'isHighCardinalityColumn' after iteration, it may not get right result with intermediate results.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -204,16 +214,17 @@ class CarbonBlockDistinctValuesCombineRDD(
...
-  extends RDD[(String, String)](prev) with Logging {
+  extends RDD[(Int, String, Boolean)](prev) with Logging {
 
-  override def getPartitions: Array[Partition] = firstParent[(Int, Array[String])].partitions
+  override def getPartitions: Array[Partition] = firstParent[(Int, ColumnDistinctValues)].partitions
 
-  override def compute(split: Partition, context: TaskContext): Iterator[(String, String)] = {
-    val LOGGER = LogServiceFactory.getLogService(this.getClass.getName)
+  override def compute(split: Partition, context: TaskContext): Iterator[(Int, String, Boolean)] = {
+    val LOGGER = LogServiceFactory.getLogService(this.getClass().getName)
```
ravipesala: remove parenthesis for this.getClass(). make like this.getClass

ravipesala:
@jarray888 Ok.
@QiangCai Please add unit test case for calculating high cardinality column. And add functional test case to check whether table is updated or not.

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
QiangCai: fixed

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonMetastoreCatalog.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: fixed in CarbonProperties

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: I think only rowCountPercentage or thresholdPercentage is not enough to identify high cardinality.
1 use rowCountPercentage: if the number of row count is small, also should use dict.
2 use highCardThreshold: if the number of row count is big enough, also should use dict.
So we need two parameters to do it.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
QiangCai: fixed

ravipesala:
@QiangCai . Please correct the test case, I will merge it.

QiangCai:
retest this please

QiangCai:
@ravipesala 
exception:
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.TimSort.mergeHi(TimSort.java:899)
	at java.util.TimSort.mergeAt(TimSort.java:516)
	at java.util.TimSort.mergeForceCollapse(TimSort.java:457)
	at java.util.TimSort.sort(TimSort.java:254)
	at java.util.Arrays.sort(Arrays.java:1438)
	at org.carbondata.processing.sortandgroupby.sortdata.SortDataRows$1.call(SortDataRows.java:367)
	at org.carbondata.processing.sortandgroupby.sortdata.SortDataRows$1.call(SortDataRows.java:1)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

solution:
If I use -Djava.util.Arrays.useLegacyMergeSort=true when starting up, no exception.
So I think there are some issues in RowComparator need to be fixed.

ravipesala:
@QiangCai Yes, there is a problem in comparsion of ByteBuffer for no dict columns. I will raise the PR . Please review and merge it. After that I will merge this PR.

ravipesala:
@QiangCai  The testcase bug is fixed in https://github.com/HuaweiBigData/carbondata/pull/523 .

QiangCai:
@ravipesala I have done rebase.

QiangCai:
@ravipesala  I have a question

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
QiangCai: I think it is better to calculoate 'isHighCardinalityColumn'  with intermediate result.

ravipesala:
@QiangCai regarding calculating highcardinality, we have 2 conditions  one is that cardinality/rowcount should be more than configurable threshold limit. For this condition rowcount and cardinality both varies if you calculate intermediate. So we get unpredictable results, better calculate after final result

QiangCai:
@ravipesala  @jarray888  If calculate after final result, why we need to submit this pr?   we need to avoid dataloading hang and OOM.

ravipesala:
@QiangCai you are right,  intermediate result is similar the first load. That is why it is a temporary solution. My opinion is at least for first load we calculate after completion of it.

QiangCai:
@ravipesala  when rowcount of intermediate is bigger than highCardThreshold, we do it. is it ok?

ravipesala:
The below is current code to calculate high cardinality dimension.

`(columnCardinality.toDouble / rowCount * 100 > model.rowCountPercentage) &&
        (columnCardinality.toDouble / model.highCardThreshold * 100 > model.thresholdPercentage)`

I think we don't require 'model.thresholdPercentage' in condition 2. You can make as follows.
`(columnCardinality.toDouble / rowCount * 100 > model.rowCountPercentage) &&
        rowCount > model.highCardThreshold `

or simply go a head with single condition 'columnCardinality > model.highCardThreshold'. In this user can configure the threshold size to treat as high cardinality column irrespective of rowCount.

Either any one of the above we can calculate in intermediate

QiangCai:
ok, change to:
(columnCardinality.toDouble / rowCount * 100 > model.rowCountPercentage) 
&& 
rowCount > model.highCardThreshold
@jarray888

ravipesala:
LGTM Merged to master

PR #399, opened 2016-05-14
Add more query in PerfTest.java

jackylk:
more queries are added: olap query, point query with primary key, detail record filter query with non-primary key, scan query with multiple sum


chenliang613:
I tested, have the error:
Point Query 2: Exception in thread "main" java.lang.RuntimeException: result is not the same between parquet and orc

jackylk:
@chenliang613 updated, it should solve your problem

chenliang613:
tested ok, merge it.

PR #402, opened 2016-05-15
Filter and encoding lookup optimizations

ravipesala:
When RLE encoding is enabled, we no need to decompress it while doing the include and exclude filters. It makes the filter query faster if we apply filters on rle data.

And also optimized the encoding lookup and used bitset instead of using list.contains.


ravipesala:
@kumarvishal09 please review the code

PR #404, opened 2016-05-16
[WIP] Optimized Filter Query and Fixed Non Filter EndKey Issue

kumarvishal09:
1)Optimized Fixed filter issue when filter is on multiple dimension
default start and key was getting created and in btree jump
it was giving all the nodes of btree
2) Now in case of filter query end key will be created based on cardinality removed getting the
max key from dictionary this is to optimize filter creation
3) Fixed No filter end key issue


sraghunandan:
Not required.Confirmed by vishal

PR #407, opened 2016-05-16
[Issue - 400] Query failed before first load

manishgupta88:
Root cause for failure:

1. No filter applied in the query -> When query was executed without data load, then number of nodes were zero and we calculate the tasks per node by dividing number of nodes with number of blocks. As a result of this Arithmetic exception (divide by zero) exception was thrown.
2. Filter applied in the query -> In this case while resolving the filter, column cache was queried to find the surrogate key and as data load did not happen the cache load was unsuccessful because of which exception was thrown


integration/spark/src/main/scala/org/carbondata/integration/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -81,43 +81,60 @@ class CarbonSparkPartition(rddId: Int, val idx: Int,
...
-    var i = 0
-    // Create Spark Partition for each task and assign blocks
-    nodeBlockMapping.asScala.foreach { entry =>
-      entry._2.asScala.foreach { blocksPerTask =>
-        if (blocksPerTask.size() != 0) {
-          result.add(new CarbonSparkPartition(id, i, Seq(entry._1).toArray, blocksPerTask))
-          i += 1;
+      // get splits
+      val splits = carbonInputFormat.getSplits(job, filterResolver);
+      val carbonInputSplits = splits.asScala.map(_.asInstanceOf[CarbonInputSplit])
```
gvramana: check case when segment exists but data doesnot exists

PR #410, opened 2016-05-16
Fix issue in create table and dictionary_include/exclue

eason-lyx:
1.create table need to handle timestamp datatype as dimension.
2.fix bug in dictionary_include/exclude

please review.
@sraghunandan @sujith71955 @Vimal-Das


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -613,13 +642,18 @@ class CarbonSqlParser()
...
+      ) {
         if (!splittedCols.isEmpty) {
-          splittedCols.foreach(dicIncludedCols =>
-            if (!field.column.equalsIgnoreCase(dicIncludedCols)) {
-              msrFields :+= field
-            }
-          )
+          breakable {
+            splittedCols.foreach(dicIncludedCols =>
+              if (!field.column.equalsIgnoreCase(dicIncludedCols)) {
```
nareshpr: In this logic, if more than 1 include columns are available, if check will fail for the first column and add it as measure.

can we convert splittedCols.toSeq/toSet and check contains ?

eason-lyx: ok,i will fix.

eason-lyx:
TestCase failure because of use the mesure to dictionary.
already fix.

QiangCai:
@eason-lyx  i think you should check with @nareshpr at first

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -541,17 +543,46 @@ class CarbonSqlParser()
     if (None != tableProperties.get("DICTIONARY_EXCLUDE")) {
       val dicExcludeCols: String = tableProperties.get("DICTIONARY_EXCLUDE").get
       splittedCols = dicExcludeCols.split(',')
+      splittedCols.foreach { splittedCol =>
+        if (!fields.map { field =>
```
gvramana: use find function Ex:
val c = ""
    fields.find( x => x.equalsIgnoreCase(c))

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -541,17 +543,46 @@ class CarbonSqlParser()
...
+            "Please check create table statement.")
+        }
+      }
     }
-    if(None != tableProperties.get("DICTIONARY_INCLUDE")) {
+
+    if (None != tableProperties.get("DICTIONARY_INCLUDE")) {
       dictIncludeCols = tableProperties.get("DICTIONARY_INCLUDE").get.split(",")
+      dictIncludeCols.foreach { distIncludeCol =>
+        if (!fields.map { field =>
```
gvramana: use find function Ex:
val c = ""
    fields.find( x => x.equalsIgnoreCase(c))

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -541,17 +543,46 @@ class CarbonSqlParser()
...
+          field.column.toLowerCase
+        }.contains(distIncludeCol.toLowerCase)) {
+          sys.error("DICTIONARY_INCLUDE column(s) is no exist in table. " +
+            "Please check create table statement.")
+        }
+      }
+    }
+
+    val dictIn = dictIncludeCols.map { dictIncludeCol =>
```
gvramana: use find function Ex:
val c = ""
    fields.find( x => x.equalsIgnoreCase(c))

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -541,17 +543,46 @@ class CarbonSqlParser()
...
+      }
     }
 
     // by default consider all String cols as dims and if any dictionary exclude is present then
     // add it to nodictionarydims list.
     fields.foreach(field => {
       if (field.dataType.get.equalsIgnoreCase("string") ||
         field.dataType.get.equalsIgnoreCase("array") ||
-        field.dataType.get.equalsIgnoreCase("struct")) {
+        field.dataType.get.equalsIgnoreCase("struct") ||
```
gvramana: Add these data types into another function, 
isDimentionDatatype(string datatype), so that directly can check in all the places, without repeating datatype names

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -613,15 +644,20 @@ class CarbonSqlParser()
...
-        !field.dataType.get.equalsIgnoreCase("struct")) {
+        !field.dataType.get.equalsIgnoreCase("struct") &&
+        !field.dataType.get.equalsIgnoreCase("timestamp")
+      ) {
         if (!splittedCols.isEmpty) {
-          splittedCols.foreach(dicIncludedCols =>
-            if (!field.column.equalsIgnoreCase(dicIncludedCols)) {
-              msrFields :+= field
+          var isDicIncludeCol = false
+          splittedCols.foreach(dicIncludedCol =>
```
gvramana: here also find can be used

eason-lyx:
all comment is handled.please check.

PR #413, opened 2016-05-16
Remove CarbonIterator

jackylk:
CarbonIterator is exactly same as java.util.Iterator, use java.util.Iterator instead


jackylk:
retest this please

kumarvishal09:
Carbon Iterator was added so no need to implement remove method, we can have abstract class which will implement iterator interface and all the concrete classes will extends the abstract class. Abstract class can have remove method implementation, so we can avoid duplicate remove method from all the concrete implementation

jackylk:
@kumarvishal09 modified

ravipesala:
@jackylk  Please rebase the PR.

jackylk:
@ravipesala rebased

core/src/main/java/org/carbondata/core/iterator/PlainIterator.java
```diff
@@ -19,20 +19,20 @@
...
-   * @return <tt>true</tt> if the iterator has more elements.
-   */
-  boolean hasNext();
+import java.util.Iterator;
+
+/**
+ * PlainIterator adds default implement for remove. This is required for Java 7.
+ * @param <E>
+ */
+public abstract class PlainIterator<E> implements Iterator<E> {
```
ravipesala: Can you name as `CarbonIterator`, so that changes will be very minimal.

ravipesala:
LGTM. Merged to master.

PR #416, opened 2016-05-17
[Issue-415] create table with DICTIONARY_EXCLUDE is not showing the excluded columns in describe

nareshpr:


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -573,18 +573,14 @@ class CarbonSqlParser()
     // add it to nodictionarydims list.
     fields.foreach(field => {
       if (isDetectAsDimentionDatatype(field.dataType.get)) {
-        var isNoDictionary = false
         if (!splittedCols.isEmpty) {
           splittedCols.foreach(excludedCol =>
```
gvramana: replace this with exists function

nareshpr:
retest this please

Issue #403, opened 2016-05-15
The building steps assume the thrift program is installed


drankye:
Following the steps to build the project, it failed in my environment due to lacking of thrift tool:

Downloaded: https://repo.maven.apache.org/maven2/com/google/collections/google-collections/0.8/google-collections-0.8.jar (473 KB at 157.1 KB/sec)
[ERROR] thrift failed output:
[ERROR] thrift failed error: 'thrift' is not recognized as an internal or external command,
operable program or batch file.

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] carbondata ......................................... SUCCESS [  5.717 s]
[INFO] carbon-common ...................................... SUCCESS [  7.693 s]
[INFO] carbon-format ...................................... FAILURE [ 20.535 s]
[INFO] carbon-core ........................................ SKIPPED
[INFO] carbon-processing .................................. SKIPPED
[INFO] carbon-hadoop ...................................... SKIPPED
[INFO] carbon-spark ....................................... SKIPPED
[INFO] carbon-assembly .................................... SKIPPED
[INFO] carbon-examples .................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 34.313 s
[INFO] Finished at: 2016-05-15T16:07:14-07:00
[INFO] Final Memory: 19M/169M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.thrift.tools:maven-thrift-plugin:0.1.11:compile (generate-thrift-java) on project carbon-format: thrift did not ex
leanly. Review output for more information. -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :carbon-format

mohammadshahidkhan:
I think we should add the building info like other open source projects are providing. For example  hadoop provides building details in 
BUILDING.txt.

drankye:
[...]

PR #429, opened 2016-05-17
add apache thrift to README.md

QiangCai:


jackylk:
But I think thrift will be downloaded and user does not need to install thrift by himself, right?

QiangCai:
@jackylk No, maven-thrift-plugin will not downlaod thirft. We need to downlad thirft and add it to path.

ravipesala:
@QiangCai  I don't think it is necessary to add any information to ReadMe file. Maven can take care of downloading and building thrift.  
@manishgupta88 @mohammadshahidkhan why it is required to add `<thriftExecutable>thrift</thriftExecutable>` to format/pom.xml. Even without this also maven is able to build format.jar file.  Please let me know so that I can raise the PR to remove it.

QiangCai:
pls have a look at [Issue-403](https://github.com/HuaweiBigData/carbondata/issues/403)

ravipesala:
@QiangCai  I am not telling it works now. But we can change the build script to make it work without installing thrift. maven will take care of compiling thrift.

chenliang613:
I agree with Ravindra's opinion.

QiangCai:
close this pr

QiangCai:
How to change the build script without installed thrift?

ravipesala:
@QiangCai  sorry, I have overseen it. Even though after I remove `<thriftExecutable>thrift</thriftExecutable> ` from pom.xml, it still looks for thrift in classpath so it is required to have thrift installed in the machine.

jackylk:
Ok. So I am merging this

PR #433, opened 2016-05-18
[Bug] Exception thrown when number of nodes less than number of blocks

manishgupta88:
Code Modifications done:
1. If number of blocks are less than number of nodes, then map of node vs block was empty and exception was thrown
2. Made default size of carbon file size to 1024
3. complex type primitive int columns are not parsed


ravipesala:
LGTM

PR #434, opened 2016-05-18
Fixed Testcase failure issue

kumarvishal09:
fixed Sum(dimension column) was failing for other data type then double


ravipesala:
LGTM

PR #436, opened 2016-05-18
Fixed length dimension aggregation issue for non number data type

kumarvishal09:
Fixed non number data type dimension column aggregation issue
@TODO need to handle for other type of dimension column , variable length, column group and complex type


ravipesala:
LGTM

Issue #450, opened 2016-05-20
Error information is not proper while loading data when the table is not exists


Jay357089:
1. lh_05  is not  exist 

then execute "load data inpath 'hdfs://hacluster/mytest/create_cube_01.csv' INTO table lh_05 OPTIONS ('DELIMITER'=',', 'QUOTECHAR'='\"');"

the error info is  "Error: org.apache.spark.sql.execution.QueryExecutionException: FAILED: ParseException line 1:79 missing EOF at 'OPTIONS' near 'lh_05' (state=,code=0)",

while the expected output should be "the table is not exist"

Jay357089:
i will raise a pr to handle this.

Jay357089:
pr merged, close it.

PR #451, opened 2016-05-20
[issue-450] Fix loading error when no table

Jay357089:
1. when loading with "OPTIONS", intercept in CarbonStrategy, do not go to hiveNativeCommand
2. change "cube" to "table" 

[Issue-450](https://github.com/HuaweiBigData/carbondata/issues/450)

@sraghunandan @gvramana @eason-lyx   pls check


QiangCai:
@Jay357089 please show me with reproduction

Jay357089:
error
correct

Jay357089:
@QiangCai  the above is the error message before and after corrected.  and i have rebased, pls check

QiangCai:
could you add test case for this issue?

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonStrategies.scala
```diff
@@ -180,8 +180,9 @@ class CarbonStrategies(sqlContext: SQLContext) extends QueryPlanner[SparkPlan] {
       case LoadCube(schemaNameOp, cubeName, factPathFromUser, dimFilesPath,
         partionValues, isOverwriteExist, inputSqlString) =>
         val isCarbonTable = CarbonEnv.getInstance(sqlContext).carbonCatalog
-          .cubeExists(schemaNameOp, cubeName)(sqlContext);
-        if (isCarbonTable) {
+          .cubeExists(schemaNameOp, cubeName)(sqlContext)
+        if (isCarbonTable ||
+          inputSqlString.split("[ ,]").exists(_.equalsIgnoreCase("OPTIONS"))) {
```
gvramana: options is optional, it cannot be mandated

Jay357089: @gvramana   maybe i do not get your point. i use "||" not "&&", so the "isCarbonTable" still works. if not use "options", then it is properly going to hiveNativeCommand. now the problem is, the carbon table is not created, neither is the hive table. but testers use "load data inpath ... options()" to test the table, and think that the error should be  about carbontable. and there is no such syntax like "load ...options" in hive, so i think can use "OPTIONS" to distinguish carbon table and hive table. pls let me know if any idea.

Jay357089:
@gvramana  pls check and give more explicit comments.

QiangCai:
**1 reproduce**
```sql
load data inpath 'hdfs://hacluster/mytest/create_cube_01.csv' INTO table lh_05 OPTIONS ('DELIMITER'=',', 'QUOTECHAR'='\"');
```
**2 exception**
 "Error: org.apache.spark.sql.execution.QueryExecutionException: FAILED: ParseException line 1:79 missing EOF at 'OPTIONS' near 'lh_05' (state=,code=0)"

this load data sql is correct for carbon, but only not exist table.  So we expect TableNotExsitExcption but find ParseException. 
I think we should fix this issue.
@gvramana  what's your idea?

Jay357089:
@gvramana   there is a DTS(DTS2016052004874) about this problem, it's urgent, so pls repsonse quickly.  @jarray888 @chenliang613

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonStrategies.scala
```diff
@@ -200,7 +200,8 @@ class CarbonStrategies(sqlContext: SQLContext) extends QueryPlanner[SparkPlan] {
         partionValues, isOverwriteExist, inputSqlString) =>
           val isCarbonTable = CarbonEnv.getInstance(sqlContext).carbonCatalog
             .cubeExists(schemaNameOp, cubeName)(sqlContext)
-          if (isCarbonTable) {
+          if (isCarbonTable ||
```
gvramana: change the check to partionValues.isEmpty, instead of doing String comparison of OPTIONS.

Jay357089:
withoptions
withoutoption

the above are the test result of with options and without options.

@QiangCai @gvramana  pls review and check in.

QiangCai:
LGTM

PR #454, opened 2016-05-20
[issue-453]Catch MalformedCarbonCommandException in dictionary_ex/in and throw directly

eason-lyx:
1.dictionary exception will be throw directly
2.hive can no paser the statement,when dictionary exception occur.


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -545,8 +549,10 @@ class CarbonSqlParser()
       splittedCols = dicExcludeCols.split(',')
       splittedCols.foreach { splittedCol =>
         if (!fields.exists(x => x.column.equalsIgnoreCase(splittedCol))) {
-          sys.error("DICTIONARY_EXCLUDE column(s) is no exist in table. " +
-            "Please check create table statement.")
+          val errormsg = "DICTIONARY_EXCLUDE column(s) is no exist in table. " +
+            "Please check create table statement."
+          logError(errormsg)
+          throw new DictionaryException(errormsg)
```
QiangCai: @eason-lyx 
1 log column info: which column is not exist
2 I don't suggest  throw new exception with same errormsg

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -555,16 +561,20 @@ class CarbonSqlParser()
       dictIncludeCols = tableProperties.get("DICTIONARY_INCLUDE").get.split(",")
       dictIncludeCols.foreach { distIncludeCol =>
         if (!fields.exists(x => x.column.equalsIgnoreCase(distIncludeCol))) {
-          sys.error("DICTIONARY_INCLUDE column(s) is no exist in table. " +
-            "Please check create table statement.")
+          val errormsg = "DICTIONARY_INCLUDE column(s) is no exist in table. " +
+            "Please check create table statement."
+          logError(errormsg)
+          throw new DictionaryException(errormsg)
```
QiangCai: @eason-lyx   need to do same thing

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -555,16 +561,20 @@ class CarbonSqlParser()
...
     }
 
     splittedCols.foreach { dicExcludeCol =>
       if (dictIncludeCols.exists(x => x.equalsIgnoreCase(dicExcludeCol))) {
-        sys.error("DICTIONARY_EXCLUDE can not contain the same column with DICTIONARY_INCLUDE. " +
-          "Please check create table statement.")
+        val errormsg = "DICTIONARY_EXCLUDE can not contain the same column " +
+          "with DICTIONARY_INCLUDE. Please check create table statement."
+        logError(errormsg)
+        throw new DictionaryException(errormsg)
```
QiangCai: @eason-lyx 
1 log column info: which column is the same column
2 I don't suggest throw new exception with same errormsg

eason-lyx:
all comment is handle.

QiangCai:
LGTM, but could you add test case to check log right ?

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonStrategy.scala
```diff
@@ -36,6 +36,10 @@ private[spark] class CarbonSQLDialect extends HiveQLDialect {
     try {
       sqlParser.parse(sqlText)
     } catch {
+      // dictionary exception need to throw directly
+      // because hive can no recognize dictionary exception
+      case de: DictionaryException =>
+        throw de
       case _ => super.parse(sqlText)
       case x: Throwable => throw x
```
jackylk: Actually this line is never hit, can we catch the all exceptions that CarbonSqlParser is going to throw in line 43, and throw the other exception here? @eason-lyx @gvramana @ravipesala @QiangCai

gvramana: We have to categorize , execptions into two categories
1. MalformedCarbonCommandException - in this case no need to fallback to hiveDialect, as this command is detected as carbon command only but malformed
2. other parse exeception - fall back to hive, as this is not Carbon Command and might be hive command.

gvramana: the same also needs to be handled, in CarbonStrategies HiveNativeCommand handling parse function

eason-lyx:
retest this please

gvramana:
retest this please

Issue #324, opened 2016-05-05
Data loading performance optimization


manishgupta88:

manishgupta88:
#444 ...PR merged

PR #455, opened 2016-05-20
[Issue-324]Data loading performance optimization: optimize FileHolderImpl

QiangCai:
1 The filePath parameter of updateCache method is  absolute path, So I think we don't need to use CarbonFile.getAbsolutePath()

2 During data loading, we invoke the updateCache method frequently, maybe reach a million times. So we should not new a instance of CarbonFile in the updataCache method.

link [Issue-324](https://github.com/HuaweiBigData/carbondata/issues/324)


jackylk:
LGTM

PR #457, opened 2016-05-20
[Issue-324]Data loading performance optimization: add MemoryMappedFileHolderImpl

QiangCai:
1 add MemoryMappedFileHolderImpl

2 use MemoryMappedFileHolderImpl in AbstractTempSortFileReader instead of FileHolderImpl

link [Issue-324](https://github.com/HuaweiBigData/carbondata/issues/324)


jarray888:
after change to map file , any performance imporve ? 
can you show some test result ?

core/src/main/java/org/carbondata/core/datastorage/store/impl/MemoryMappedFileHolderImpl.java
```diff
@@ -0,0 +1,103 @@
+package org.carbondata.core.datastorage.store.impl;
```
jackylk: please add license header

core/src/main/java/org/carbondata/core/datastorage/store/impl/MemoryMappedFileHolderImpl.java
```diff
@@ -0,0 +1,103 @@
...
+public class MemoryMappedFileHolderImpl implements FileHolder {
+
+  private static final LogService LOGGER =
+      LogServiceFactory.getLogService(MemoryMappedFileHolderImpl.class.getName());
+
+  private Map<String, FileChannel> fileNameAndStreamCache;
+  private Map<String, MappedByteBuffer> fileNameAndMemoryMappedFileCache;
+
+  public MemoryMappedFileHolderImpl() {
+    this.fileNameAndStreamCache =
```
jackylk: just call MemoryMappedFileHolderImpl(int capacity) instead of writing it again

core/src/main/java/org/carbondata/core/datastorage/store/impl/MemoryMappedFileHolderImpl.java
QiangCai: fix

core/src/main/java/org/carbondata/core/datastorage/store/impl/MemoryMappedFileHolderImpl.java
QiangCai: fixed

jackylk:
LGTM

PR #425, opened 2016-05-17
[Issue-328]  Data load failing when the measure data is null

foryou2030:
fixed issue #328 
1 change calculateMaxMinUnique method name to calculateMaxMin 
2 null value don't need to call calculateMaxMin. fixed the load exception.

what's more, null measure value was stored as 0 , maybe null value should be stored as unique value.

PR #466, opened 2016-05-20
[Issue- 408] Data Load is failing when no measure is used in create cube command

mohammadshahidkhan:
Loading data with measure column was failing, to fix the same added added the dummy measure when no measure is defined the create table DDL.

Select * from TableName;
Output
image

Must be merged after PR  #425

### DIFF
--- a/format/src/main/thrift/schema.thrift
+++ b/format/src/main/thrift/schema.thrift
@@ -85,6 +85,10 @@ struct ColumnSchema{
 	11: optional string aggregate_function;
 
 	12: optional binary default_value;
+	/**
+	* To specify the visibily of the column by default its false
+	*/
+	13: optional bool invisible;
 }
 
 /**
--- a/core/src/main/java/org/carbondata/core/carbon/metadata/schema/table/column/ColumnSchema.java
+++ b/core/src/main/java/org/carbondata/core/carbon/metadata/schema/table/column/ColumnSchema.java
@@ -100,6 +100,11 @@ public class ColumnSchema implements Serializable {
   private byte[] defaultValue;
 
   /**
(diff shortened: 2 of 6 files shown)


gvramana:
retest this please

format/src/main/thrift/schema.thrift
```diff
@@ -85,6 +85,10 @@ struct ColumnSchema{
 	11: optional string aggregate_function;
 
 	12: optional binary default_value;
+	/**
+	* To specify the visibily of the column by default its true
+  **/
+	13: required bool visible = true;
```
gvramana: Move required field to invisible and make in optional

format/src/main/thrift/schema.thrift
```diff
@@ -85,6 +85,10 @@ struct ColumnSchema{
 	11: optional string aggregate_function;
 
 	12: optional binary default_value;
+	/**
+	* To specify the visibily of the column by default its true
+  **/
```
jackylk: incorrect indentation

format/src/main/thrift/schema.thrift
jackylk: Agree to @gvramana and I think we should build mechanism for backward compatibility for any change in carbon-format layer, especially adding new **required** field

mohammadshahidkhan:
Fixed Review comments
Build successful with running 277 Testcase
image

gvramana:
Add a check in Carbon Relation to skip invisible columns.

mohammadshahidkhan:
Fixed review comment 
please review the same

ravipesala:
retest this please

gvramana:
retest this please

PR #468, opened 2016-05-21
[BUG] Add exception handle in MDKeyGenStep

Jay357089:


Jay357089:
@QiangCai pls check

processing/src/main/java/org/carbondata/processing/mdkeygen/MDKeyGenStep.java
```diff
@@ -188,13 +188,18 @@ public boolean processRow(StepMetaInterface smi, StepDataInterface sdi) throws K
...
+      LOGGER.error(e, "Failed for table: " + this.tableName + " in MDKeyGenStep");
       throw new KettleException("Error while initializing data handler : " + e.getMessage());
+    } catch (Exception e) {
+      LOGGER.error(e, "Failed for table: " + this.tableName + " in MDKeyGenStep");
+      throw new KettleException("There is an unexpected error: " + e.getMessage());
     } finally {
       try {
         dataHandler.finish();
       } catch (CarbonDataWriterException e) {
         LOGGER.debug("Error in  closing data handler ");
```
QiangCai: I prefer to use error method instead of debug metod

Jay357089: @QiangCai it is the original code. i don't know why they use LogDebug, but Are you serious that we can change?  @gvramana

jackylk: I think change to log.error is required. This is not a proper use of debug log

Jay357089:
@QiangCai @jackylk  pls review and check in.

PR #475, opened 2016-05-23
[ISSUE-474]fix bug in load data with exclude_dictionary and empty dimensions

eason-lyx:
in surrogate generate step,when change null value to null string ,currently using [meta.memberMapping.length] which is dim.length+measure.length to loop the row.
but row size = dim.length+measure.length+nodictionary.length.
currently process logic will result in null point exception when using dictionary_exclude.
change using row size to loop the row.


jackylk:
@eason-lyx can you add a test case for it using dictionary_exclude

eason-lyx:
retest this please

eason-lyx:
@jackylk testcase is added.
please review and check in.

integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataload/TestDataWithDicExcludeAndInclude.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  override def beforeAll {
+    buildTestData
+    buildTable
+  }
+
+  test("test load data with dictionary exclude & include and with empty dimension") {
+    loadTable
```
jackylk: please do a query to test the table is correctly loaded and compare the query test against the expected result.

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -833,8 +833,6 @@ private String getCarbonLocalBaseStoreLocation() {
         if (obj.equals(valueToCheckAgainst)) {
           rowValue[i] = CarbonCommonConstants.MEMBER_DEFAULT_VAL;
         }
-      } else if (i > meta.memberMapping.length) {
```
gvramana: 1. valueToCheckAgainst is never initialized.
2. This comparision applies to only dimension fields  i.e.,dictionary+nondictionary+complextypes_members
@kumarvishal09 @nareshpr please comment on this

eason-lyx: @gvramana @kumarvishal09 @nareshpr 
1.is valueToCheckAgainst alway use null?
2.the comparision i think we applies to all fields.
   1)if we give only part of fileds,we will get null when processing surrogate step.
   2)meausre is null ,we set to "@NU#LL$!" here. and when handle measure column at following process,we will set measure to null if the value of measure is "@NU#LL$!".and according the datatype give the measure 0L or 0.0.

eason-lyx:
retest this please

integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataload/TestDataWithDicExcludeAndInclude.scala
eason-lyx: ok,already added a query.

gvramana:
retest this please

PR #478, opened 2016-05-24
[Issue- 476] data load is failing with empty dimension column 

mohammadshahidkhan:
data load is failing with empty dimension column


processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -884,7 +884,7 @@ private String getCarbonLocalBaseStoreLocation() {
     for (int j = 0; j < inputColumnsSize; j++) {
       String columnName = metaColumnNames[j];
       String foreignKeyColumnName = foreignKeyMappingColumns[j];
-      r[j] = ((String) r[j]).trim();
+      r[j] = null != r[j] ? ((String) r[j]).trim() : CarbonCommonConstants.MEMBER_DEFAULT_VAL;
```
eason-lyx: i think no need to assigh the null string value to r[j] again.
it should be handle by one place->changeNullValueToNullString()
i have fix this in pr#475. u can check. :)

mohammadshahidkhan:
Closing this PR will be fixed with PR #475

PR #480, opened 2016-05-24
[Issue- 479] Direct dictionary dataload failure when used with No dictionary columns

mohammadshahidkhan:
1. Direct dictionary dataload failure when used with No dictionary columns
2. Clean up is not happening after the TestTimestampDataTypeDirectDictionary execution
3. Fixed Import issue of CarbonMetadataUtil
4. Rename TimestampDataTypeDirectDictionary_FT to TestTimestampDataTypeDirectDictionary


ravipesala:
@mohammadshahidkhan Please add testcase for this scenerio

mohammadshahidkhan:
@ravipesala Added Test case for the mentioned scenario
@gvramana Handled the review comments as per our discussion to pass the serialized string of the ColumnSchemaDetailsWrapper format 
Please review the same

mohammadshahidkhan:
Fixed UT failure 
was falling due to different case file header.
ColumnSchemaDetailsWrapper .get(columnName) was returning null due different case name .
Now doing map population and lookup with columnId instead of column name

gvramana:
LGTM

gvramana:
retest this please

PR #481, opened 2016-05-25
[Issue-83] Make invertedIndex can be configurable

Zhangshunyu:
Using ddl to determine which column should not use inverted index like this：

```
 (ID Int, date Timestamp, country String,
name String, phonetype String, serialname String, salary Int)
TBLPROPERTIES('NO_INVERTED_INDEX'='ID,name')

LOAD DATA LOCAL INPATH './src/test/resources/invertedIndex.csv' into table index
```
By default, we use inverted index if the user didn't set this for one column.

### DIFF
--- a/core/src/main/java/org/carbondata/core/keygenerator/directdictionary/DirectDictionaryGenerator.java
+++ b/core/src/main/java/org/carbondata/core/keygenerator/directdictionary/DirectDictionaryGenerator.java
@@ -40,4 +40,17 @@ public interface DirectDictionaryGenerator {
    */
   Object getValueFromSurrogate(int key);
 
+  /**
+   * The method generate and returns the dictionary / surrogate key for direct dictionary column
+   * This Method is called while executing filter queries for getting direct surrogate members.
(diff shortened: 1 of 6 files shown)


Vimal-Das:
Can you share performance Impact of disabling inverted index
1) Is there improvement in data loading speed?
2) Is there compression ratio improvement/degrade?
3) Is there difference in query performance?

Vimal-Das:
My understanding of issue 83 was to add a TBLProperties in DDL to allow power user to configure the inverted index at table and column level.

gvramana:
Yes, it should be configurable at table and column level.

gvramana:
Please contact @ashokblend to discuss about column properties, whic he has introduced

Zhangshunyu:
@gvramana @Vimal-Das OK, I will do it as you recommented.

Zhangshunyu:
retest this please

Zhangshunyu:
@Vimal-Das @gvramana pls review this code, thank you.

Zhangshunyu:
retest this please

Zhangshunyu:
Could any committer help to review this code?

PR #487, opened 2016-05-25
[Issue-486] Segment merge causing the IOException

### LINKED ISSUE #486: Segment merge causing the IOException
currently segment merge is not supported as per new store format
 disable the segment merge feature

mohammadshahidkhan:
Fixed #486 
currently segment merge is not supported as per new store format
disabling the segment merge feature


ravipesala:
LGTM

PR #497, opened 2016-05-27
Instead of fixed way, make the cores used for block sort confiugrable.

Zhangshunyu:
The user can configure this num according to the configuration of the machine, and sometimes if the machine still has enough cores after exclude "carbon.number.of.cores.while.loading", we can set this value "carbon.number.of.cores.block.sort" to larger to avoid RejectedExecutionException, but if the machine has not enough cores we can reduce it.


Zhangshunyu:
@Vimal-Das @QiangCai @manishgupta88 pls review this code, thank you!

Vimal-Das:
LGTM, But at the same time I am wondering whether we can avoid introducing one more configurable property.

Can we reuse the MDK sort thread count property as when block sort is happening, MDK sort is already completed, can we keep the thread count same as MDK sort?

Zhangshunyu:
@Vimal-Das I think that while final merger read rows from files, the Producer will generate mdk and try to start block sort thread who is responsible for RLE and inverted index, meanwhile the Consumer will write the blocklet files, they are concurrent streaming like procedure. Here the "block sort " means RLE and inverted index.

processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -651,7 +655,14 @@ private NodeHolder getNodeHolderObject(byte[][] dataHolderLocal, byte[][] byteAr
         }
       }
     }
-    ExecutorService executorService = Executors.newFixedThreadPool(7);
+    try {
+      thread_pool_size = Integer.parseInt(CarbonProperties.getInstance()
+              .getProperty(CarbonCommonConstants.NUM_CORES_BLOCK_SORT,
+                      CarbonCommonConstants.NUM_CORES_BLOCK_SORT_DEFAULT_VAL));
+    } catch (NumberFormatException e) {
+      thread_pool_size = Integer.parseInt(CarbonCommonConstants.NUM_CORES_BLOCK_SORT_DEFAULT_VAL);
```
QiangCai: better to validate property in CarbonProperties

Zhangshunyu:
@Vimal-Das Hi vimal, what's your opinion?

Zhangshunyu:
@ravipesala pls check this code，thank you.

processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -651,7 +655,14 @@ private NodeHolder getNodeHolderObject(byte[][] dataHolderLocal, byte[][] byteAr
...
+    }
```
ravipesala: Please move these validations to CarbonProperties class

Zhangshunyu:
@ravipesala Done, pls check again.

QiangCai:
LGTM. merge to master.

PR #503, opened 2016-05-27
[issue-502] Dictionary exclude column should be dimension not measure

ManoharVanam:
Dictionary exclude column will be dimension


integration/spark/src/test/scala/org/apache/spark/sql/TestCarbonSqlParser.scala
```diff
@@ -76,14 +76,290 @@ class TestCarbonSqlParser extends QueryTest {
...
     //No dictionary column names will be available in noDictionary list
+    assert(noDictionary.size == 1)
     assert(noDictionary.lift(0).get.equalsIgnoreCase("col2"))
   }
 
+  test("Test-DimAndMsrColsWithNoDictionaryFields1") {
+    val tableProperties = Map("DICTIONARY_EXCLUDE" -> "col1")
+    var fields: Seq[Field] = Seq[Field]()
+
+    var col1 = Field("col1", Option("Int"), Option("col1"), None, null, Some("columnar"))
```
gvramana: move common part to methods

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -637,24 +638,26 @@ class CarbonSqlParser()
...
-    var splittedCols: Array[String] = Array[String]()
+    var dictIncludedCols: Array[String] = Array[String]()
+    var dictExcludedCols: Array[String] = Array[String]()
 
-    // check any column need to be excluded
+    // get all included cols
     if (None != tableProperties.get("DICTIONARY_INCLUDE")) {
-      val dicIncludeCols: String = tableProperties.get("DICTIONARY_INCLUDE").get
-      splittedCols = dicIncludeCols.split(',')
+      dictIncludedCols = tableProperties.get("DICTIONARY_INCLUDE").get.split(',')
```
eason-lyx: i think we need to add trim for include/exclude cols.

ManoharVanam: yeah , need to add the same in the old flow also

integration/spark/src/test/scala/org/apache/spark/sql/TestCarbonSqlParser.scala
ManoharVanam: ok

ravipesala:
LGTM

PR #508, opened 2016-05-28
[Issue - 506] CarbonSqlParser is not validating the duplicate column …

mohammadshahidkhan:
…name if they are in different case in create table command

Modification:
Now in case duplicate columns  system will throw error while creating table.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -362,7 +362,18 @@ class CarbonSqlParser()
         case list@Token("TOK_TABCOLLIST", _) =>
           val cols = BaseSemanticAnalyzer.getColumns(list, true)
           if (cols != null) {
-
+            val dupColsGrp = cols.asScala
+              .groupBy(x => x.getName) filter { case (_, colList) => colList
+              .size > 1
+            }
+            if (dupColsGrp.size > 0) {
+              var columnName: String = ""
+              dupColsGrp.toSeq.foreach(columnName += _._1 + ", ")
+              columnName = columnName.substring(0, columnName.lastIndexOf(", "))
+              val errorMessage = "Duplicate column name: " + columnName + " found in table " +
+                ".Please check create table statement."
+              throw new MalformedCarbonCommandException(errorMessage)
+            }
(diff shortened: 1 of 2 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -362,7 +362,10 @@ class CarbonSqlParser()
         case list@Token("TOK_TABCOLLIST", _) =>
           val cols = BaseSemanticAnalyzer.getColumns(list, true)
           if (cols != null) {
-
+            var dup = cols.asScala.groupBy(x => x) filter { case (_, colList) => colList.size > 1 }
+            if (dup.size > 0) {
```
nareshpr: column name is same, datatype is different, will this fix work ?

mohammadshahidkhan:
Fixed review comments

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
mohammadshahidkhan: handled now comparing on the column name

ravipesala:
@mohammadshahidkhan  Please rebase the code

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -362,7 +362,18 @@ class CarbonSqlParser()
         case list@Token("TOK_TABCOLLIST", _) =>
           val cols = BaseSemanticAnalyzer.getColumns(list, true)
           if (cols != null) {
-
+            val dupColsGrp = cols.asScala
+              .groupBy(x => x.getName) filter { case (_, colList) => colList
```
ravipesala: groupBy is not case sensitive right? How the duplicate columns with different case letters are grouped together?

mohammadshahidkhan: val cols = BaseSemanticAnalyzer.getColumns(list, true) is returning the cols in lower case only so no need to handle the case.

mohammadshahidkhan:
rebase done

ravipesala:
LGTM

PR #512, opened 2016-05-28
[Issue 513]Add tableId as part of CarbonTableIdentifier

ashokblend:
CarbonTableIdentifier have only databaseName and tableName. 
Add tableId also as part of CarbonTableIdentifier.

### DIFF
--- a/core/src/main/java/org/carbondata/core/carbon/CarbonTableIdentifier.java
+++ b/core/src/main/java/org/carbondata/core/carbon/CarbonTableIdentifier.java
@@ -37,11 +37,17 @@ public class CarbonTableIdentifier implements Serializable {
   private String tableName;
 
   /**
+   * table unique id
+   */
+  private int tableId;
+
+  /**
    * constructor
    */
-  public CarbonTableIdentifier(String databaseName, String tableName) {
+  public CarbonTableIdentifier(String databaseName, String tableName, int tableId) {
     this.databaseName = databaseName;
     this.tableName = tableName;
+    this.tableId = tableId;
   }
 
   /**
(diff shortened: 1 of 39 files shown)


core/src/main/java/org/carbondata/core/carbon/CarbonTableIdentifier.java
```diff
@@ -37,11 +37,17 @@
...
   /**
+   * table unique id
+   */
+  private int tableId;
+
+  /**
    * constructor
    */
-  public CarbonTableIdentifier(String databaseName, String tableName) {
+  public CarbonTableIdentifier(String databaseName, String tableName, int tableId) {
```
ravipesala: Please add one more constructor with only `databaseName` and `tableName`. Other wise it would be lot of unncessary changes.

gvramana: @ravipesala tableId is always required, to identify a table uniquely.So it becomes mandatory parameter.
But for external users who directly want to use, tableidentifier, then we have to give interface to get tableIdentifier given databasename and tablename

ravipesala: @gvramana Why tableId is mandatory to use? table can be uniquely identified by dbname and table name. It cannot be allowed to create duplicate tables anyway. I am sorry if I am out of sync.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -336,7 +336,7 @@ class TableNewProcessor(cm: tableModel, sqlContext: SQLContext) {
     val schemaEvol = new SchemaEvolution()
     schemaEvol
       .setSchemaEvolutionEntryList(new util.ArrayList[SchemaEvolutionEntry]())
-    tableSchema.setTableId(1)
+    tableSchema.setTableId(scala.util.Random.nextInt())
```
ravipesala: how random could be unique?

jackylk: You can refer to [this](http://www.javapractices.com/topic/TopicAction.do?Id=56) to generate unique ID

Vimal-Das: Yes, we should use UUID instead of random number.
But that would require changing the tableID to string.
In that case we can rename the variable in thrift also to tableUUID

ashokblend:
another PR is created for this #614 . Closing this PR.

PR #515, opened 2016-05-28
[Issue-514] When struct data type data has more number of values then desired columns, throws indexOutOfBound

nareshpr:


processing/src/main/java/org/carbondata/processing/datatypes/StructDataType.java
```diff
@@ -173,7 +173,7 @@ public void parseStringAndWriteByteArray(String tableName, String inputString,
       dataOutputStream.writeInt(children.size());
       delimiterIndex =
           (delimiter.length - 1) == delimiterIndex ? delimiterIndex : delimiterIndex + 1;
-      for (int i = 0; i < splitInput.length; i++) {
+      for (int i = 0; i < splitInput.length && i < children.size(); i++) {
```
ravipesala: what if the columns are less? Does it treat as null?
Please add one test case for columns are less scenario as well

nareshpr: for children without any data, below for loop will handle treating it as null.

Already test data (complexdata.csv) is having case where null and less column data for struct is available.

Testcases are available @ TestLoadDataWithHiveSyntax.test("complex types data loading")  for (mobile struct<imei:string, imsi:string>) column data in complexdata.csv

ravipesala:
retest this please

ravipesala:
LGTM

PR #518, opened 2016-05-28
[Issue 517]fixed group by count query

ashokblend:
Carbon is assuming below query as count query, which is not correct.
select ,count() from table group by having count()>1

Fix is made to do proper check for count query.


ravipesala:
retest this please

ravipesala:
Please add test case for this scenerio

integration/spark/src/main/scala/org/apache/spark/sql/CarbonOperators.scala
```diff
@@ -534,6 +534,21 @@ case class CarbonCubeScan(
...
 
+  /**
+   * return true if query is count query
+   * @return
+   */
+  def isCountQuery: Boolean = {
+    if (buildCarbonPlan.isCountStarQuery() && null == buildCarbonPlan.getFilterExpression &&
+        buildCarbonPlan.getDimensions.size() < 1 && buildCarbonPlan.getMeasures.size() < 2 &&
+        buildCarbonPlan.getDimAggregatorInfos.size() < 1 &&
+        buildCarbonPlan.getExpressions.size() == 0) {
```
ravipesala: I am doubtful about this check `buildCarbonPlan.getExpressions.size()`. No where we are adding expressions in our code.

gvramana:
retest this please

Issue #529, opened 2016-05-30
Error info is not proper when content of input file is empty


Jay357089:
1. Prepare a null csv file  
  the csv file has nothing.
2. Create table:
CREATE TABLE group06 (imei string,age int,productdate timestamp,gamePointId double) STORED BY 'org.apache.carbondata.format'  TBLPROPERTIES('COLUMN_GROUPS'='(imei,age)');
3.Load the csv file to the table:
LOAD DATA INPATH 'hdfs://hacluster/data/null.csv'  INTO TABLE group06 options ('DELIMITER'=',', 'QUOTECHAR'='\"', 'FILEHEADER'= 'imei,age,productdate,gamePointId');

then dataload failed, and the error info is not explicit.

Jay357089:
i will raise a pr for this

Jay357089:
pr merged. close it.

PR #530, opened 2016-05-30
[issue-529] Fix bug of loading when input file exists but has nothing.

Jay357089:
issue #529 

1. add logWarning for empty content file 
2. add logWarning for non csv file
3. throw exception when input files are all empty content or not csv format


ravipesala:
@Jay357089 Please add test case for the scenario mentioned in issue https://github.com/HuaweiBigData/carbondata/issues/529. I will merge it.

Jay357089:
@ravipesala handled. pls check

Jay357089:
@QiangCai pls review.

Issue #531, opened 2016-05-30
The error of '/ by zero' will random occurs when query from the table.


mohammadshahidkhan:
'/ by zero ' could come in the scenario when the ('int noOfTasksPerNode = parallelism / mapOfNodes.size();' )  noOfTasksPerNode is zero.

The noOfTasksPerNode  can return 0 when parallelism is less the mapOfNodeds.size()

mohammadshahidkhan:
pr merged closing the issue

PR #532, opened 2016-05-30
[Issue - 531] The error of '/ by zero' will random occurs when query from the table

mohammadshahidkhan:
#531 , modification
Assigning at least 1 task per node, instead of zero.


ravipesala:
LGTM merged to master

PR #536, opened 2016-05-30
[Performance Improvement] Data loading performance optimization

manishgupta88:
Code Modifications done:
1. Surrogate key gen Step: Modified code to get max cardinality of a column from dictionary cache and not read the dictionary metadata file again and again.
2. Sort Key step: 
a. Implemented producer and consumer and moved sorting and writing in different thread which will optimize the writing speed of data.
b. Above change will also help in reducing GC as this step was one of the root cause of GC thus increasing the data loading time.
c. Removed unsued code while writing data to a sort temp file.


kumarvishal09:
retest this please

kumarvishal09:
retest this please

PR #537, opened 2016-05-30
Fixed performance issue in data load

kumarvishal09:
Fixed data load performance issue. 
1) for each cell value we were comparing no dictionary columns with all the columns to check whether column is dictionary column or not
2) Avoid no dictionary byte array copy.now directly wrapping to byte buffer

### DIFF
--- a/processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
+++ b/processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
@@ -135,14 +135,7 @@ public class CarbonCSVBasedSeqGenStep extends BaseStep {
    * Csv file path
    */
   private String csvFilepath;
-  /**
-   * modifiedDimesions
-   */
-  private String modifiedDimesions;
-  /**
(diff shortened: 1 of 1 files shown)


processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1734,14 +1735,10 @@ private void clearDictionaryCache() {
     }
   }
 
-  private void processnoDictionaryDim(int index, String dimension, ByteBuffer[] out) {
-
-    String dimensionValue = dimension;
-    ByteBuffer buffer = ByteBuffer.allocate(dimensionValue.length());
-    buffer.put(dimensionValue.getBytes(Charset.forName("UTF-8")));
+  private void processnoDictionaryDim(int index, String dimensionValue, ByteBuffer[] out) {
+    ByteBuffer buffer = ByteBuffer.wrap(dimensionValue.getBytes(Charset.forName("UTF-8")));
```
ravipesala: Use CommonConstant for UTF-8

gvramana:
retest this please

gvramana:
retest this please

PR #545, opened 2016-05-31
[issue-553][issue-544]fix anyorder in schema with complex datatype

eason-lyx:
currently, we assume complex datatype should be at last.otherwise,exception occur.
for this pr:
1.fix any order in schema with complex datatype
2.modify some no correct parameter

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -481,9 +481,8 @@ class CarbonSqlParser()
     val partitioner: Option[Partitioner] = getPartitionerObject(partitionCols, tableProperties)
 
     if (!checkComplexDimPos(dims)) {
-      val errorMsg = "Complext data type column should be at the end of dimensions, " +
-        "please check the create statement."
-      throw new MalformedCarbonCommandException(errorMsg)
+      throw new MalformedCarbonCommandException("Complext data type column should be at the end" +
+        " of dimensions, please check the create statement.")
     }
 
     tableModel(ifNotExistPresent,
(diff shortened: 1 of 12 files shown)


eason-lyx:
@ravikiran23 @nareshpr please help to review.

core/src/main/java/org/carbondata/core/carbon/metadata/schema/table/column/ColumnSchema.java
```diff
@@ -105,6 +105,11 @@
   private boolean invisible = false;
 
   /**
+   * parent column schema
+   */
+  private ColumnSchema parentColumnSchema;
```
nareshpr: Why parentColumnSchema is required in childColumnSchema?
Thrift is not having parentColumnSchema place holder. Reloading table metadata to catalog may have some impact.

Better to handle columns reordering in CarbonSqlParser before calling Create after preparing table column list.

QiangCai:
add links of issue
[issue-553](https://github.com/HuaweiBigData/carbondata/issues/553) Dataload need to sort order:dimension+complex+measure
[issue-544](https://github.com/HuaweiBigData/carbondata/issues/544) Data load failure when statement contain complex datatype and dictionary_include contain double type

eason-lyx:
all comment handled.please review.
@nareshpr

eason-lyx:
after discuss with @kumarvishal09 ,suggestion is when complex datatype is on at the end of dimension,then throw MalformedCarbonCommandException.

processing/src/main/java/org/carbondata/processing/graphgenerator/GraphGenerator.java
```diff
@@ -67,7 +67,16 @@
 
 public class GraphGenerator {
 
+  /**
+   * global block info
+   */
   public static final HashMap<String, BlockDetails[]> blockInfo = new HashMap<>();
```
kumarvishal09: better use Map<String,BlockDetails[]> blockInfo = new HashMap<>()

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -473,6 +479,26 @@ class CarbonSqlParser()
   }
 
   /**
+   * check the dimensions whether complex type at last
+   * @param dims
+   * @return
+   */
+  protected def checkComplexDimPos(dims: Seq[Field]): Boolean = {
+    val complexDims: ArrayBuffer[(Field, Integer)] = new ArrayBuffer()
```
kumarvishal09: only position you can store in the array buffer no need to store field

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -465,6 +465,12 @@ class CarbonSqlParser()
 
     val partitioner: Option[Partitioner] = getPartitionerObject(partitionCols, tableProperties)
 
+    if (!checkComplexDimPos(dims)) {
+      val errorMsg = "Complext data type column should be at the end of dimensions, " +
+        "please check the create statement."
+      throw new MalformedCarbonCommandException(errorMsg)
```
QiangCai: no need to define variable, please use directly

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -473,6 +479,26 @@ class CarbonSqlParser()
...
+  protected def checkComplexDimPos(dims: Seq[Field]): Boolean = {
+    val complexDims: ArrayBuffer[(Field, Integer)] = new ArrayBuffer()
+    val otherDims: ArrayBuffer[(Field, Integer)] = new ArrayBuffer()
+    dims.zipWithIndex.foreach { dim =>
+      dim._1.dataType.getOrElse("NIL") match {
+        case "array" =>
+          complexDims.+=((dim._1, dim._2))
+        case "struct" =>
+          complexDims.+=((dim._1, dim._2))
+        case _ => otherDims.+=((dim._1, dim._2))
```
QiangCai: suggest to remove point

QiangCai:
better to find root cause and fix it.

eason-lyx:
retest this please

eason-lyx:
retest this please

PR #547, opened 2016-05-31
Simplify the exception handle in loadCube

Jay357089:
the original handling is a little bit awkward. When user uses beeline, he may just gets the error info "Dataload failure. Please check the logs " in client, which is not good, and should make user get the specific error info directly.


Jay357089:
@ravipesala pls check.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1475,9 +1475,8 @@ private[sql] case class LoadCube(
       }
       catch {
         case ex: Exception =>
-          LOGGER.error(ex)
           LOGGER.audit("Dataload failure. Please check the logs")
-          sys.error("Dataload failure. Please check the logs")
+          sys.error(ex.getMessage)
```
ravipesala: What if the exception is thrown? Will it show complete stack trace on console?

Jay357089:
retest this please

Jay357089:
@ravipesala  i checked in beeline, it's OK. complete stack trace will not show on the console, they will just show in the log.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1477,7 +1477,7 @@ private[sql] case class LoadCube(
         case ex: Exception =>
           LOGGER.error(ex)
           LOGGER.audit("Dataload failure. Please check the logs")
-          sys.error("Dataload failure. Please check the logs")
+          sys.error(ex.getMessage)
```
ravipesala: What I meant is , Even if we log like ` sys.error(ex.getMessage)`, we may not get useful information without log stack trace. So user need to look in the logs to find the reason for failure anyway. 
Or else we can throw exception here so that the stack trace would be shown on console.

Jay357089: @ravipesala usually,we throw exception like this "throw new Exception("Exmessage")", and the user can get the "Exmessage" on the console by using sys.error(ex.getMessage)  and "Exmessage" should be enough for the common user. but if we give the whole stack trace to user on the console, i think it is too much. what is your idea?

jackylk: I think same logic apply to console also, user may want to see the stack trace in the console so he can understand what happened

Jay357089:
@ravipesala @jackylk  handled.

ravipesala:
LGTM

Issue #550, opened 2016-06-01
[Bug]The value of "multiline" get from partionValues map during data loading should ignore case 


Zhangshunyu:
If the user set the value of multiline as "TRUE" in UpperCase in LOAD ddl, it will be treated as false.

PR #551, opened 2016-06-01
[issue-550]Verify the boolen value of multiline

Zhangshunyu:
#550 
1.If the user set the value of multiline as "TRUE" in UpperCase in LOAD ddl, it will be treated as false, which is not correct.
2.If the user set it to "Tru" or "abc", it will be set to false without syntax check, this is not reasonable, we should check the syntax and tell the user his mistake.


Zhangshunyu:
@ravipesala pls check.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1413,11 +1413,11 @@ private[sql] case class LoadCube(
       val quoteChar = partionValues.getOrElse("quotechar", "\"")
       val fileHeader = partionValues.getOrElse("fileheader", "")
       val escapeChar = partionValues.getOrElse("escapechar", "")
-      val multiLine = partionValues.getOrElse("multiline", false)
+      val multiLine = partionValues.getOrElse("multiline", "false")
       val complex_delimiter_level_1 = partionValues.getOrElse("complex_delimiter_level_1", "\\$")
       val complex_delimiter_level_2 = partionValues.getOrElse("complex_delimiter_level_2", "\\:")
       var booleanValForMultiLine = false
```
jackylk: I think we can simply do 
```
val multiLine = partionValues.getOrElse("multiline", "false").equalsIgnoreCase("true")
```

Zhangshunyu: OK， I will do it.

Zhangshunyu:
@jackylk Done, pls check.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1413,13 +1413,10 @@ private[sql] case class LoadCube(
       val quoteChar = partionValues.getOrElse("quotechar", "\"")
       val fileHeader = partionValues.getOrElse("fileheader", "")
       val escapeChar = partionValues.getOrElse("escapechar", "")
-      val multiLine = partionValues.getOrElse("multiline", false)
+      val booleanValForMultiLine = partionValues.getOrElse("multiline", "false").
```
jackylk: why not just keep the same variable name, `multiLine` is more meaningful

Zhangshunyu:
@jackylk OK, done, pls check again.

QiangCai:
LGTM

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1413,13 +1413,10 @@ private[sql] case class LoadCube(
       val quoteChar = partionValues.getOrElse("quotechar", "\"")
       val fileHeader = partionValues.getOrElse("fileheader", "")
       val escapeChar = partionValues.getOrElse("escapechar", "")
-      val multiLine = partionValues.getOrElse("multiline", false)
+      val multiline = partionValues.getOrElse("multiline", "false").
```
jackylk: please put the `.` to next line

Zhangshunyu:
@jackylk Done.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1413,13 +1413,10 @@ private[sql] case class LoadCube(
       val quoteChar = partionValues.getOrElse("quotechar", "\"")
       val fileHeader = partionValues.getOrElse("fileheader", "")
       val escapeChar = partionValues.getOrElse("escapechar", "")
-      val multiLine = partionValues.getOrElse("multiline", false)
+      val multiline = partionValues.getOrElse("multiline", "false")
+        .equalsIgnoreCase("true")
```
QiangCai: suggest to validate multiline.
if multiline not in ("true", "false") , it is invalid.

Zhangshunyu:
@QiangCai Done, pls check.

QiangCai:
Verified. LGTM

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -1413,13 +1414,18 @@ private[sql] case class LoadCube(
...
       val escapeChar = partionValues.getOrElse("escapechar", "")
-      val multiLine = partionValues.getOrElse("multiline", false)
+      val multiline = partionValues.getOrElse("multiline", "false")
       val complex_delimiter_level_1 = partionValues.getOrElse("complex_delimiter_level_1", "\\$")
       val complex_delimiter_level_2 = partionValues.getOrElse("complex_delimiter_level_2", "\\:")
       var booleanValForMultiLine = false
-      if (multiLine.equals("true")) {
-        booleanValForMultiLine = true
-      }
+        try {
```
jackylk: wrong indentation

Zhangshunyu:
@jackylk Because the value of multiLine is set in LOAD DDL by the user and we can't ensure that the his input is legal, for example, if the user's input is "tru" but not 'true', we should use try block to check the legality, the logic in "toBoolean" will check this, so we need this try block.If it is illegal input we should tell the user their mistake.
Further more, in the previous implementation, we just simply do as this:
`val multiLine = partionValues.getOrElse("multiline", "false").equalsIgnoreCase("true")`
if the input is "tru" I mean the user's real intention is set true, but we process it as false in this logic, I think it is not unreasonable.

jackylk:
Yes, I got your point. What I am suggesting is to do a check without introducing try block. like 
```
      val multiLine = partionValues.getOrElse("multiline", "false") match {
        case "true" => true
        case "false" => false
        case other => throw new IllegalArgumentException("xxx")
      }
```

Zhangshunyu:
@jackylk Done, pls check again.

jackylk:
Thanks for working on this

PR #555, opened 2016-06-01
[Issue-556] Support for bigInt and decimal data type

manishgupta88:
Code Modifications done:
1. Modified code to parse decimal and bigInt datatype in create table command (handled by @nareshpr ).
2. Modified data load flow to load data with big decimal and bigInt data type.
3. Added test cases for testing decimal and bigInt datatype functionality.

Usage example:
Create table tablename (col1 String, col2 decimal(9,2)) stored by 'org.apache.carbondata.format'
Create table tablename (col1 String, col2 bigInt) stored by 'org.apache.carbondata.format'


core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -2,39 +2,75 @@
...
-        return Double.valueOf(msrValue);
+  /**
+   * This method will convert a given value to its specific type
+   *
+   * @param msrValue
+   * @param dataType
+   * @param carbonMeasure
+   * @return
+   */
+  public static Object getMeasureValueBasedOnDataType(String msrValue, String dataType,
```
ravipesala: Pass dataType enum instead of String. Better make consistant in all places. 
And also switch case instead of if else.
Even in `CarbonCSVBasedSeqGenStep` use DataType enum.

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -2,39 +2,75 @@
...
   public static char getAggType(String dataType) {
-    switch (dataType) {
-      case "Decimal":
-        return CarbonCommonConstants.BIG_DECIMAL_MEASURE;
-      case "BigInt":
-      case "Long":
-        return CarbonCommonConstants.BIG_INT_MEASURE;
-      default:
-        return 'n';
+    if (DataType.DECIMAL.name().equals(dataType)) {
```
ravipesala: Pass dataType enum and use switch case

core/src/main/java/org/carbondata/query/aggregator/impl/DistinctCountBigDecimalAggregatorObjectSet.java
```diff
@@ -0,0 +1,151 @@
...
+
+  public DistinctCountBigDecimalAggregatorObjectSet() {
+    valueSetForObj = new HashSet<Object>(CarbonCommonConstants.DEFAULT_COLLECTION_SIZE);
+  }
+
+  /**
+   * just need to add the unique values to agg set
+   */
+  @Override public void agg(double newVal) {
+    valueSetForObj.add(newVal);
```
ravipesala: I don't think this case can come. Better don't implement it.

core/src/main/java/org/carbondata/query/aggregator/impl/DistinctCountLongAggregatorObjectSet.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.carbondata.core.carbon.datastore.chunk.MeasureColumnDataChunk;
+import org.carbondata.core.constants.CarbonCommonConstants;
+import org.carbondata.query.aggregator.MeasureAggregator;
+
+public class DistinctCountLongAggregatorObjectSet implements MeasureAggregator {
```
ravipesala: Most of the methods are implemented in same way for 'DistinctCountBigDecimalAggregatorObjectSet' and `DistinctCountLongAggregatorObjectSet` classes. I can see `agg(MeasureColumnDataChunk dataChunk, int index)` and `getNew' methods are deffered. So better extract Abstract class and extend them. And also find any similar classes can come under this abstract class

core/src/main/java/org/carbondata/query/aggregator/impl/AbstractMeasureAggregatorMaxMin.java
```diff
@@ -21,7 +21,6 @@
```
ravipesala: The abstract class is not abstracting methods properly. I guess the methods `writeData`, `getByteArray`, `readData`, `merge` have same implementations across all sub classes. So better move them to this abstract class. And also create new abstract sub classes separately for  Max type and min type aggregator and move `internalAgg` method to it.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -337,6 +337,14 @@ class CarbonSqlParser()
         }
     }
 
+  private def getScaleAndPrecision(dataType: String): (Int, Int) = {
```
ravipesala: What if user don't provide scale and precision. Is it already handled by hive parser. What are the default values?

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -379,12 +387,17 @@ class CarbonSqlParser()
               val dataType = Option(col.getType)
               val name = Option(col.getName())
               // This is to parse complex data types
-              val f: Field = anyFieldDef(new lexical.Scanner(col.getName + ' ' + col.getType))
+              var f: Field = anyFieldDef(new lexical.Scanner(col.getName + ' ' + col.getType))
```
ravipesala: why it is changed to `var`. It supposed be `val`

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -379,12 +387,17 @@ class CarbonSqlParser()
...
                 match {
                 case Success(field, _) => field
                 case failureOrError => new Field(columnName, dataType, name, None, null,
                   Some("columnar"))
               }
+              if("decimal".equalsIgnoreCase(f.dataType.getOrElse(""))) {
```
ravipesala: NIT. space after if

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -379,12 +387,17 @@ class CarbonSqlParser()
               val dataType = Option(col.getType)
               val name = Option(col.getName())
               // This is to parse complex data types
-              val f: Field = anyFieldDef(new lexical.Scanner(col.getName + ' ' + col.getType))
+              var f: Field = anyFieldDef(new lexical.Scanner(col.getName + ' ' + col.getType))
                 match {
```
ravipesala: Please format code properly, there is no proper indentation

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -337,6 +337,14 @@ class CarbonSqlParser()
         }
     }
 
+  private def getScaleAndPrecision(dataType: String): (Int, Int) = {
+    var splittedColGrps: Seq[String] = Seq[String]()
```
ravipesala: This variable is never used

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -69,7 +69,8 @@ case class tableModel(
 
 case class Field(column: String, dataType: Option[String], name: Option[String],
     children: Option[List[Field]], parent: String = null,
-    storeType: Option[String] = Some("columnar"))
+    storeType: Option[String] = Some("columnar"),
+    var precision: Int = 0, var scale: Int = 0)
```
ravipesala: I think these are supposed to be `val`

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -147,7 +148,7 @@ class TableNewProcessor(cm: tableModel, sqlContext: SQLContext) {
         encoders.add(Encoding.DICTIONARY)
         val coloumnSchema: ColumnSchema = getColumnSchema(
```
ravipesala: NIT correct the `coloumnSchema`

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -147,7 +148,7 @@ class TableNewProcessor(cm: tableModel, sqlContext: SQLContext) {
...
           normalizeType(field.dataType.getOrElse("")), field.name.getOrElse(field.column), index,
-          isCol = true, encoders, isDimensionCol = true, rowGroup)
+          true, encoders, true, rowGroup, field.precision, field.scale)
```
ravipesala: Format is not proper and also name the Boolean parameters like ` isCol = true`.

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
```diff
@@ -185,12 +188,8 @@ class TableNewProcessor(cm: tableModel, sqlContext: SQLContext) {
       val encoders = new java.util.ArrayList[Encoding]()
       encoders.add(Encoding.DICTIONARY)
       val columnSchema: ColumnSchema = getColumnSchema(normalizeType(field.dataType.getOrElse("")),
-        field.name.getOrElse(field.column),
-        index,
-        isCol = true,
-        encoders,
-        isDimensionCol = true,
-        -1)
+        field.name.getOrElse(field.column), index, true, encoders, true, -1,
```
ravipesala: Format is not proper and also name the Boolean parameters like `isCol = true`

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
manishgupta88: If user dont provide scale and precision, its already ahndled same as hive and default values are (10,0)

integration/spark/src/main/scala/org/apache/spark/sql/execution/command/carbonTableSchema.scala
nareshpr: By default scale & precision will be 0, we are re-assigning the values in parser only if decimal datatype. otherwise, default values remain as is.

manishgupta88:
@ravipesala ...I have handled all the review comments. Kindly review and merge

ravipesala:
LGTM

ravipesala:
retest this please

PR #557, opened 2016-06-01
Record performance statistics during data loading

Zhangshunyu:
1. After data loaded successfully, the main statistics information will be printed by each executor
2. At the same time, some tiny info like write one temp sort file/compress mdk data time/write block.. time will be printed immediately in log.

I just take CarbonExample who loads only 1000 lines data as an example, you can see the statistics info as following in executor side:
```
]==================== DummyLoadRDD Cost Time: 5.085(s)
]============================== lru Cache Load Cost Time: 0.024(s)
]==========**********Compress One Node For One Thread pool-24-thread-1 Cost: 0.026
]=====##### Write One Node Cost Time: 0.0#####=====
]========== TIME STATISTICS PartitionID: 0==========
]STATISTICS ->Raw data IO cost(load csv to dataFrame and generate block dict): 0.132(s)
]STATISTICS ->Distinct Value IO cost(global dict shuffle and write dict file): 0.648(s)
]STATISTICS ->  |_maximum distinct column shuffle time: 0.162(s)
]STATISTICS ->  |_maximum distinct column write dict file time: 0.35(s)
]STATISTICS ->Total cost of gen surrogate key, sort and write to temp files: 0.522(s)
]STATISTICS ->  |_read cost of raw csv file: 0.305(s)
]STATISTICS ->  |_cost of transform to surrogate key: 0.485(s)
]STATISTICS ->  |_io cost(sort rows and write to temp file): 0.417(s)
]STATISTICS ->IO cost(tansform to MDK, compress and write fact file): 0.629(s)
]==============================
]========== BLOCK_INFO ==========
]BLOCK_INFO ->Node host: localhost
]BLOCK_INFO ->The block count in this node: 1
]==============================
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,9 @@
+# This github has migrated to apache: https://github.com/apache/incubator-carbondata, please fork new github.
 # CarbonData
 CarbonData is a new Apache Hadoop native file format for faster 
 interactive query using advanced columnar storage, index, compression 
 and encoding techniques to improve computing efficiency, in turn it will 
 help speedup queries an order of magnitude faster over PetaBytes of data. 
-
-### Why CarbonData
-Based on the below requirements, we investigated existing file formats in the Hadoop eco-system, but we could not find a suitable solution that can satisfy all the requirements at the same time,so we start designing CarbonData. 
-* Requirement1:Support big scan & only fetch a few columns 
-* Requirement2:Support primary key lookup response in sub-second. 
-* Requirement3:Support interactive OLAP-style query over big data which involve many filters in a query, this type of workload should response in seconds. 
-* Requirement4:Support fast individual record extraction which fetch all columns of the record. 
-* Requirement5:Support HDFS so that customer can leverage existing Hadoop cluster. 
 
 ### Features
 CarbonData file format is a columnar store in HDFS, it has many features that a modern columnar format has, such as splittable, compression schema ,complex data type etc. And CarbonData has following unique features:


jackylk:
can you take a look a #356  to check whether you can leverage the Profiler there?

Zhangshunyu:
@This pr has no impact on pr 356.

Zhangshunyu:
@jackylk This pr has no impact on pr 356.

jackylk:
I mean can you re-use the profiler in pr356. pr356 has a profiler that you can define your load phase. We can use the same to record performance for both query and load

kumarvishal09:
@Zhangshunyu please check the performance impact after you change

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsUtil.java
```diff
@@ -0,0 +1,294 @@
...
+          Collections.synchronizedSortedSet(new TreeSet<Long>());
+  //Due to thread thread blocking in each task, we only record the max
+  //dicWriteFile Time of each single thread
+  private static SortedSet<Long> dicWriteFileMaxTime =
+          Collections.synchronizedSortedSet(new TreeSet<Long>());
+  private static SortedSet<Long> dicShuffleAndWriteFileTotalTime =
+          Collections.synchronizedSortedSet(new TreeSet<Long>());
+
+  //Generate surrogate keys total time:
+  private static ConcurrentHashMap<String, SortedSet<Long>> parSurrogatekeysTotalTimeMap =
```
kumarvishal09: private static Map<String, SortedSet<Long>> parCsvInputStepTimeMap =
 +          new ConcurrentHashMap<String, SortedSet<Long>>();

kumarvishal09:
for data loading and query statistics we can have one common interface, which will have multiple implementation, default will be dummy which will not record any thing . And one property we can expose like carbon.record.statistics. factory will return the concrete implementation based on the property value. in normal case it will record dummy so it will not impact the performance

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsUtil.java
```diff
@@ -0,0 +1,294 @@
...
+import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.carbondata.common.logging.LogService;
+import org.carbondata.common.logging.LogServiceFactory;
+
+/**
+ * A util which provide a method used to count time information.
+ */
+public class CarbonTimeStatisticsUtil {
```
kumarvishal09: there are so many map and set used in this class, why this is required?

core/src/main/java/org/carbondata/core/util/CarbonTimeStatisticsUtil.java
```diff
@@ -0,0 +1,294 @@
...
+
+  private static final LogService LOGGER =
+          LogServiceFactory.getLogService(CarbonTimeStatisticsUtil.class.getName());
+  //Generate global dictionay total time :
+  private static SortedSet<Long> loadCsvfilesToDfTime =
```
kumarvishal09: how many values we are going to store here in all the map and sets, i think only max value,then in that case we can have one class which will be used to store the max value .

PR #561, opened 2016-06-01
[Issue 560] Zookeeper test case fails randomly and generates big log files under core package

ravipesala:
There are three issues with that ZooKeeper test case
1. Some times test case hangs while starting the zookeeper server, that is the bug in zookeeper 3.4.8, it could be solved by either downgrading to 3.4.7 or overriding the ZooKeeperServer start method. Please have a look at https://issues.apache.org/jira/browse/ZOOKEEPER-2383

2. It generates very big log files under core/version-2 folder. It could be solved by moving the folder to target.

3. Test case some times failing because it cannot get the node, It is because the baseNode existance and creation is added to Watcher. It calls every time for each call. Calls some times hangs there. So we can pre-create the base node solves the issue.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -172,7 +172,7 @@
     <dependency>
       <groupId>org.apache.zookeeper</groupId>
       <artifactId>zookeeper</artifactId>
-      <version>3.4.8</version>
+      <version>3.4.7</version>
     </dependency>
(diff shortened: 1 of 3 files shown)


jackylk:
LGTM. @gvramana please check also

Issue #566, opened 2016-06-03
load data failed when using complextype  in a different order


foryou2030:
create statement:
`create table arraytableprop20 (deviceInformationId int, channelsId string, ROMSize string, mac array<string>, purchasedate string, gamePointId double, contractNumber double) STORED BY 'org.apache.carbondata.format'`
   
load statement
   
`LOAD DATA INPATH 'D:/hadoop/srcdata/Array_2.csv' INTO table arraytableprop20 options ('DELIMITER'=',', 'QUOTECHAR'='"', 'FILEHEADER'='deviceInformationId,channelsId,ROMSize,purchasedate,mac,gamePointId,contractNumber','COMPLEX_DELIMITER_LEVEL_1'='$')`

load data failed and catch exception:
```
WARN  03-06 11:19:36,095 - Removing executor driver with no recent heartbeats: 252130 ms exceeds timeout 120000 ms
ERROR 03-06 11:20:09,951 - Lost executor driver on localhost: Executor heartbeat timed out after 252130 ms
ERROR 03-06 11:20:09,956 - [arraytableprop20: Graph - Carbon Surrogate Key Generator][partitionID:0] 
java.lang.RuntimeException: java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer
	at org.carbondata.processing.surrogatekeysgenerator.csvbased.CarbonCSVBasedSeqGenStep.process(CarbonCSVBasedSeqGenStep.java:831)
	at org.carbondata.processing.surrogatekeysgenerator.csvbased.CarbonCSVBasedSeqGenStep.processRow(CarbonCSVBasedSeqGenStep.java:435)
	at org.pentaho.di.trans.step.RunThread.run(RunThread.java:50)
	at java.lang.Thread.run(Thread.java:722)
Caused by: java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer
	at org.carbondata.processing.util.RemoveDictionaryUtil.prepareOut(RemoveDictionaryUtil.java:50)
	at org.carbondata.processing.surrogatekeysgenerator.csvbased.CarbonCSVBasedSeqGenStep.populateOutputRow(CarbonCSVBasedSeqGenStep.java:1144)
	at org.carbondata.processing.surrogatekeysgenerator.csvbased.CarbonCSVBasedSeqGenStep.process(CarbonCSVBasedSeqGenStep.java:820)
	... 3 more
```

PR #567, opened 2016-06-03
[Issue-566] copy out to newArray should use memberMapping[index]

foryou2030:
in this pr, i have fixed the exception in #566 
When do populateOutputRow, we copy the "out" value to "newArray" finnally.
In this function(prepareOut) , we shuld using the same index as out array, so memberMapping[index] is proper


foryou2030:
@QiangCai pls check

PR #569, opened 2016-06-03
[Issue-339] Block balance for data load

gvramana:
Data load distribution logic for block locality
    1) Blocks load distribution based on block locality
    2) Blocks size reduction to parallelize block reading by logically
    splitting bigger blocks to small blocks. This will be done only, if not sufficient blocks exists for parallelism.


core/src/main/java/org/carbondata/core/carbon/path/CarbonTablePath.java
```diff
@@ -170,6 +170,19 @@ public String getCarbonDataFileName(Integer filePartNo, Integer taskNo,
         + CARBON_DATA_EXT;
   }
 
+  /**
+   * check if it is carbon data file matching extension
+   * @param fileNameWithPath
+   * @return boolean
+   */
+  public static boolean isCarbonDataFile(String fileNameWithPath) {
+    int pos = fileNameWithPath.lastIndexOf('.');
```
kumarvishal09: we can directly check filepath ends with CARBON_DATA_EXT.

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -1117,27 +1118,34 @@ private static void assignLeftOverBlocks(Map<String, List<TableBlockInfo>> outpu
   private static void createOutputMap(Map<String, List<TableBlockInfo>> outputMap,
       int blocksPerNode, Set<TableBlockInfo> uniqueBlocks,
       Map<String, List<TableBlockInfo>> nodeAndBlockMapping) {
+
+    ArrayList<NodeMultiBlockRelation> multiBlockRelations =
```
kumarvishal09: Any specific reason to use arraylist type reference??Use Generic List reference instead of array list reference to hold arraylist object.

integration/spark/src/main/scala/org/apache/spark/rdd/DummyLoadRDD.scala
```diff
@@ -56,10 +56,8 @@ class DummyLoadRDD(prev: NewHadoopRDD[LongWritable, Text])
       override def next(): (String, BlockDetails) = {
```
kumarvishal09: Are we using this class? If not can we delete??

core/src/main/java/org/carbondata/core/carbon/path/CarbonTablePath.java
gvramana: no, later wants to add_extension for update

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
gvramana: this is temporary/local variable, so fine

integration/spark/src/main/scala/org/apache/spark/rdd/DummyLoadRDD.scala
gvramana: ok

gvramana:
@kumarvishal09 fixed comments

kumarvishal09:
retest this please

PR #422, opened 2016-05-17
Filter optimization with range index in fwd dictionary cache lookup

sujith71955:
Below implementation is been done for providing optimization in forward dictionary cache look up while applying filters, a range based index is provided for faster look up so that always the dictionary look up will start based on provided range

PR #570, opened 2016-06-03
[#389] Added range level index while reading the forward dictionary c…

sujith71955:
This feature will perform range based scanning on forward dictionary cache, while performing binary search for retrieving the dictionary value of filter member the e system memorizes the previous range up to which the scanning performed.

This feature can improve the performance of 'IN' filters.


sujith71955:
retest this please

sujith71955:
PR #422 is same to this PR

PR #573, opened 2016-06-04
Optimized detail query flow and push up computation flow.

ravipesala:
1. Detail query flow no longer uses aggregate objects.
2. Enabled prefetch for raw record query.
3. Removed unnecessary code from result preparator for detail query flow.

Improvement:
Old code for detail scan query
Scan Query 0: 917, 754, **1921** [sql: Select sum(c7), sum(c8), avg(c9), max(c10) from tableName ]

After optimizations:
Scan Query 0: 854, 755, **1036** [sql: SELECT sum(c7), sum(c8), avg(c9), max(c10) FROM tableName ]


integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -134,10 +135,10 @@ class CarbonQueryRDD[K, V](
   }
 
 
-  override def compute(thepartition: Partition, context: TaskContext): Iterator[(K, V)] = {
+  override def compute(thepartition: Partition, context: TaskContext): Iterator[V] = {
     val LOGGER = LogServiceFactory.getLogService(this.getClass.getName)
-    val iter = new Iterator[(K, V)] {
-      var rowIterator: CarbonIterator[RowResult] = _
+    val iter = new Iterator[V] {
+      var rowIterator: CarbonIterator[Array[Object]] = _
```
gvramana: Array[Object] would have wrapped with RawRow

ravipesala:
Rebased

core/src/main/java/org/carbondata/query/carbon/result/iterator/DetailRawQueryResultIterator.java
```diff
@@ -48,26 +56,63 @@ public DetailRawQueryResultIterator(List<BlockExecutionInfo> infos,
...
-    } else {
-      return queryResultPreparator.prepareQueryResult(null);
+    return result;
+  }
+
+  private ResultInfo getResultFromFuture(Future<ResultInfo> future) {
+    try {
+      return future.get();
+    } catch (Exception e) {
+      e.printStackTrace();
```
gvramana: throw when exception

PR #574, opened 2016-06-04
[Issue 548] Fixed all find bugs in all modules

ravipesala:
This PR fixes all find bug issues in all modules


hadoop/src/main/java/org/carbondata/hadoop/util/ObjectSerializationUtil.java
```diff
@@ -33,18 +37,29 @@ public static String convertObjectToString(Object obj) throws IOException {
...
+          oos.close();
+        }
+        if (gos != null) {
+          gos.close();
+        }
+        if (baos != null) {
+          baos.close();
+        }
+      } catch (IOException e) {
+        LOG.error(e);
```
jackylk: should throw the exception again, right?

ravipesala: Throwing the exception in finally block is not recommendable as it confuses the user and masks the actual exceptions. In two cases the flow comes to finally block
 First case is after successful completion of flow in try block and reaches finally block , in this case even though the exception the thrown while closing the connections cannot effect the flow, so we just log it.
Second case is something abnormal happened in try block flow and throws exception, so before throwing exception it goes to finally block to close the connections. In this case if we throw exception from finally block , it masks the actual exception.  That is why throwing exception in finally block is not recommended.

gvramana:
@ravipesala Please rebase

ravipesala:
@gvramana Rebased. Please review and merge

jackylk:
LGTM

PR #580, opened 2016-06-06
[Bug] Handle clearing of access count logic for dictionary column eviction policy

manishgupta88:
Code Modifications done:
1. Modified code to invoke clear method implementation from dictionary cache. Whenever a query is fired and dictionary cache is getting used, we should take copy and continue. This is required for column to be successfully deleted from LRU cache.
2. Fixed LRU cache issue in which data load and query was executed successfully in case lru cache size was set to 0MB. If LRU cache size is set to zero then default behavior needs to be considered and all columns based on request should be loaded in memory i.e no eviction policy.


Vimal-Das:
I suggest that we keep track of access count for correctly identifying the memory used by dictionary so that we can correctly adhere to the user configuration of restricting the dictionary cache to a fixed limit.

manishgupta88:
@Vimal-Das ...I have handled the review comments. Kindly review and merge

Vimal-Das:
retest this please

Vimal-Das:
LGTM

Vimal-Das:
Thanks for working on this, Merged to master.

PR #594, opened 2016-06-07
[issue-593] Fixed delete load by id command issue

ManoharVanam:
1. Delete segment without segment id should throw exception
2. using segment load time from tablestatus file instead of using thread sleep


Vimal-Das:
retest this please

Vimal-Das:
LGTM

Vimal-Das:
Merged to master.

PR #596, opened 2016-06-07
[Issue 595]Pushdown Equijoin to carbon

ashokblend:
1. Modified CarbonDataSourceRelation to get relation size(table size)
3. Modified CarbonStrategies@canpushdown to allow pushdown in case of both table are carbontable
3. Renamed FilterPushJoin to BroadcastFilterPushJoin
4. Modified BroadcastFilterPushJoin to broadcast small table data to executors

TO-DO
Need to check table size computable for large dataset


integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
```diff
@@ -132,6 +134,14 @@ private[sql] case class CarbonDatasourceRelation(
   def schema: StructType = carbonRelation.schema
 
   def sqlContext: SQLContext = context
+
+  override def sizeInBytes: Long = {
+     val tablePath = carbonRelation.cubeMeta.dataPath + CarbonCommonConstants.FILE_SEPARATOR +
+     carbonRelation.cubeMeta.carbonTableIdentifier.getDatabaseName +
+     CarbonCommonConstants.FILE_SEPARATOR +
+     carbonRelation.cubeMeta.carbonTableIdentifier.getTableName
+    FileUtils.getSpaceOccupied(tablePath)
```
ravipesala: Can you get the time of calculating size of table from hdfs? it would be called many times. It should not impact the actual performance

ravipesala: better change `def` to `val`. So that it would be calculated once for query, otherwise it would try to calculate for each call

Vimal-Das: getSpaceOccupied will return the total space, including the space required for replication, can you check.

ashokblend:
@ravipesala 
I checked for 1.35Mb data and it is checking 1ms. I will check for large dataset

@Vimal-Das 
I have added FileFactory@getDirectorySize : for hdfs it will give size without considering replication.
hdfs filesystem api provides content summary which gives both size with replication and without replication.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
gvramana: giving val can be problem, when relation is created and it will be cached, if dataloading happens later this will not be considered. So we can check metafile time stamp and reload if time stamp changes.

Issue #581, opened 2016-06-06
LOAD DDL could not use keywords of carbon's datatype to define a table name, which is incompatible with HIVE.


Zhangshunyu:
For example:
`create table _timestamp_ (imei string,AMSize string,channelsId string,ActiveCountry string, Activecity string,gamePointId double,deviceInformationId double,productionDate Timestamp,deliveryDate timestamp,deliverycharge double) STORED BY 'org.apache.carbondata.format'TBLPROPERTIES('DICTIONARY_INCLUDE'='gamePointId,deviceInformationId');

LOAD DATA INPATH 'hdfs://hacluster/Raghu/vardhandaterestruct.csv' INTO TABLE _timestamp_ OPTIONS('DELIMITER'=',', 'QUOTECHAR'= '"', 'FILEHEADER'= 'imei,deviceInformationId,AMSize,channelsId,ActiveCountry,Activecity,gamePointId,productionDate,deliveryDate,deliverycharge');

Error: org.apache.spark.sql.AnalysisException: missing EOF at 'OPTIONS' near 'timestamp'; line 1 pos 87 (state=,code=0)

`
timetamp is a keyword, and if the table name is a keyword, it would be parsed failure.

foryou2030:
Another question, if we create a table with name "table_03" and load data with name "Table_03", load failed.  This is because carbon is case sensitive. But hive is case insensitive.
Shall we keep the same with hive?

@gvramana @kumarvishal09 @QiangCai pls check.

Zhangshunyu:
#597

Zhangshunyu:
#704 
this pr fix this issue.

PR #597, opened 2016-06-07
[issue-581]fix the bug of load parse failure when tblname is in carbon's datatype keywords in DDL.

Zhangshunyu:
#581 
1.Solve this bug : #581 
Separate different ddl and dml, which can seperate the keywords and fix the bug of load parse failure when tblname is carbon's datatype keywords in DDL. Actually, all the parse about tablename.Further more, if the ddl is not related to table name, it can still use the CarbonParser to parse.
2.Use base/child class to prevent keyword from over coupling without changing the parse logic.

We introduced so many key words, for example, table name is : "NO_DICTIONARY", it will be parsed by CREATE DDL because we use hive to parse it, but it could not be parsed by LOAD DDL because in our parser it is a key word. This is not reasonable.


Zhangshunyu:
@ravipesala pls review this code, thank you!

gvramana:
I think these changes are not required, can explain the error, for which this is fixed.

Zhangshunyu:
for example，if the tablename is string/timestamp/no_dictionary…etc，the tablename will be treated as keywords in cafbonparser，and many ddl will be failure.this is incompetable with hive，we should keep the key words seperated to avoid this.

Zhangshunyu:
you can see the testcase i provided

gvramana:
Not required to move to base class, initLexical already will have keywords, needs to validate wrt that

Zhangshunyu:
@gvramana Pls talk with sujith，he will explain this.

Zhangshunyu:
Now，our parser can not parse this ddl：
LOAD DATA INPATH './src/test/resources/tblNameWithKeywordLoad.csv' INTO TABLE
 STRING OPTIONS('DELIMITER'=',', 'QUOTECHAR'= '"')
The table name is STRING，I mean table name not data type

Vimal-Das:
@Zhangshunyu I had a discussion with @gvramana , we think that we should further investigate whether there is any difference when validating the keywords between Hive Parser and Spark Parser.
We need to make both Create Cube and Load Data DDL have consistent behavior.

If I create a Table with Name as Load, Create DDL will succeed, but loading data will fail. This is not expected behaviour. Either Create should fail or create and load both should suceed.

Zhangshunyu:
Maybe we should keep some reserved keywords，who apperenced in BaseCarbonSqlParser

Zhangshunyu:
What should we do now? Our load ddl it different from hive, not like create table DDL.
@Vimal-Das @gvramana 
I mean, now how to handle the DTS?

Zhangshunyu:
@gvramana @sujith71955 
I tested using sparksql, I think sparksql also has this problem,  pls check the following example:
```
> create table cache(a string);
+---------+--+
| result  |
+---------+--+
+---------+--+
No rows selected (0.078 seconds)
```
t> 
```
cache table cache;
Error: org.apache.spark.sql.AnalysisException: cannot recognize input near 'cache' 'table' 'cache'; line 1 pos 0 (state=,code=0)
```

PR #601, opened 2016-06-07
[Bug]Fixed multiple aggregation on same dimesion column issue

kumarvishal09:
Fixed multiple aggregation on same dimension column issue in query plan we were adding list for each aggregation type


sujith71955:
retest this please

PR #608, opened 2016-06-08
[Issue-553][Issue-544] Complex type data loading is failing if column order is not same as table order in csv file.

nareshpr:
1) Complex type data loading is failing if column order is not same as table order in csv file.
2) Removed unused code
3) Added reordering code for complex type columns.


processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenMeta.java
```diff
@@ -373,6 +376,14 @@ public void setComplexTypes(Map<String, GenericDataType> complexTypes) {
     this.complexTypes = complexTypes;
   }
```
gvramana: Use index based lookup for mapping, Using map will slow down dataload

nareshpr:
retest this please

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1038,108 +1062,105 @@ private String getCarbonLocalBaseStoreLocation() {
...
+              ByteBuffer.wrap(byteArray.toByteArray());
+          if (null != byteArray) {
+            byteArray.close();
           }
-          i++;
-        } else {
-          Dictionary dicCache = dictionaryCaches.get(foreignKeyColumnName);
+        } catch (IOException e1) {
+          throw new KettleException(
+              "Parsing complex string and generating surrogates/ByteArray failed. ", e1);
```
gvramana: I think record fetching should translate into bad record or null

PR #609, opened 2016-06-08
[Bug 610]carbondatasourcerelation size issue fix

ashokblend:
When table is dropped and user runs below command
"drop table if exists <tablename>"
carbon is giving issue while evaluating size as directory doesn't exist.

Fixe is made to check if directory exists than only evaluate size.


ravipesala:
LGTM

PR #611, opened 2016-06-08
tableblockinfo sorting was wrong.

ravikiran23:
// segment id is changed from int to string.
// changed != to equalignorecase.

### DIFF
--- a/core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
+++ b/core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
@@ -122,7 +122,7 @@ public class TableBlockInfo implements Serializable, Comparable<TableBlockInfo> 
       return false;
     }
     TableBlockInfo other = (TableBlockInfo) obj;
-    if (!segmentId.equalsIgnoreCase(other.segmentId)) {
+    if (!segmentId.equals(other.segmentId)) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
```diff
@@ -122,7 +122,7 @@ public long getBlockLength() {
       return false;
     }
     TableBlockInfo other = (TableBlockInfo) obj;
-    if (segmentId != other.segmentId) {
+    if (!segmentId.equalsIgnoreCase(other.segmentId)) {
```
jackylk: segmentId should be case sensitive, so should use `equals` only

Vimal-Das: as @jackylk mentioned, segment ID should be case sensitive, Is there any specific reason for doing case insensitive comparison?

ravikiran23: segment id is not given by user. system will give the segment id. segment id will start from 0 and will increment by 1.   in case of compaction it will be 0.1 , 0.2 etc.

Vimal-Das:
LGTM, Merged to master.

PR #613, opened 2016-06-09
[Bug] Handled Clearing access count for a dictionary column in compaction and limit case

manishgupta88:
Code Modifications done:
1. Modified code to handle clearing of access count in compaction operation and query with limit case.
2. Corrected bug in forward and reverse dictionary to clear dictionary access count in case of failure to load dictionary.


Vimal-Das:
LGTM, Merged to master.

PR #614, opened 2016-06-09
[Issue 513]Add tableId as part of CarbonTableIdentifier

ashokblend:
1.Changed tableId to string from int in schema.thrift
2.Added tableId in CarbonTableIdentifier
3.Made CarbonTable as hub to get carbonTableIdentifier or absoluteTableIdentifier

Impact: Since thrift file is changed for schema. Hence old cube will not work. We need to provide utility to change old schema as per this change.


ashokblend:
retest this please

ashokblend:
retest this please

PR #615, opened 2016-06-09
[Issue - 612] Direct dictionary improvement 1. Remove property "carbo…

mohammadshahidkhan:
…n.direct.dictionary" to enable/disable the direct dictionary feature.

even though user configures timestamp type column to be NO_DICTIONARY col, treat that column as Direct dictionary column


integration/spark/src/main/java/org/carbondata/spark/readsupport/SparkRowReadSupportImpl.java
```diff
@@ -34,6 +35,14 @@
             break;
           default:
         }
+      } else if (carbonColumns[i].hasEncoding(Encoding.DIRECT_DICTIONARY)) {
+        //convert the long to timestamp in case of direct dictionary column
+        switch (carbonColumns[i].getDataType()) {
+          case TIMESTAMP:
```
Vimal-Das: Why do you need a switch case for just one case handling?

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -724,7 +728,9 @@ class CarbonSqlParser()
     fields.foreach(field => {
 
       if (dictExcludeCols.toSeq.exists(x => x.equalsIgnoreCase(field.column))) {
-        noDictionaryDims :+= field.column
+        if (DataTypeUtil.getDataType(field.dataType.get.toUpperCase()) != DataType.TIMESTAMP) {
```
Vimal-Das: I think, we should still support Timestamp to be no dictionary, if user prefers that.
This is just in case, the data is in different format compared to the configured date time format.
@gvramana What is your opinion?

mohammadshahidkhan: @Vimal-Das @gvramana In case of NO DICTIONARY while loading we are converting the date string to binary value and writting the same in the fact, but while quering we need the timestamp 
format since we are converting the data based on the datatype
DataTypeUtil.getDataBasedOnDataType(String data, DataType actualDataType)

integration/spark/src/main/java/org/carbondata/spark/readsupport/SparkRowReadSupportImpl.java
mohammadshahidkhan: fixed converted the switch to if

Vimal-Das:
retest this please

Vimal-Das:
@mohammadshahidkhan , build is failing. Please check

Vimal-Das:
retest this please

Vimal-Das:
@mohammadshahidkhan -42 tests in carbon-spark, can you please check.

mohammadshahidkhan:
Fixed #672

Vimal-Das:
retest this please

core/src/main/java/org/carbondata/query/aggregator/impl/AvgTimestampAggregator.java
```diff
@@ -0,0 +1,114 @@
...
+        firstTime = false;
+      }
+      return;
+    } else if (newVal instanceof Timestamp) {
+      aggVal += ((Timestamp) newVal).getTime();
+      count++;
+      firstTime = false;
+      return;
+    }
+    aggVal += ((Number) newVal).doubleValue();
```
Vimal-Das: add if check to verify if its instance of Number, 

Also avoid adding returns in between, instead use if-else-if in subsequent code to avoid entering the code.

mohammadshahidkhan: fixed

Vimal-Das:
LGTM, Merged to master.

PR #617, opened 2016-06-09
Added validation for checking duplicate dictionary include/ exclude fields

ManoharVanam:
Added validation for checking duplicate dictionary include/ exclude fields


Vimal-Das:
If a same column is provided more than once in either DICTIONARY_INCLUDE or DICTIONARY_EXCLUDE, consider it only once and continue. No need for fail the create cube ddl.

ManoharVanam:
This validation is not required

PR #623, opened 2016-06-10
Fixed Issues regarding Filter, There were issues in dictionary search

sujith71955:
Fixed Issues regarding Filter, There were issues in dictionary search  for range filters if filter members contains Null.

Also handled the scenario where unsupported data type is been returned by Spark


sujith71955:
retest this please

core/src/main/java/org/carbondata/core/iterator/CarbonIterator.java
```diff
@@ -19,20 +19,17 @@
...
+ *
  * @param <E>
  */
-public abstract class CarbonIterator<E> implements Iterator<E> {
-
-  @Override public abstract boolean hasNext();
+public abstract class CarbonIterator<E> {
 
-  @Override public abstract E next();
+  public abstract boolean hasNext() throws QueryExecutionException;
```
ravipesala: It is not recommendable to throw exception in `hasNext`. Just return false if any exception occurs and log it. Please check the documentation of `java.util.Iterator<E>`

core/src/main/java/org/carbondata/core/iterator/CarbonIterator.java
```diff
@@ -19,20 +19,17 @@
...
 
-  @Override public void remove() {
-    throw new UnsupportedOperationException("remove");
-  }
+  public abstract E next() throws QueryExecutionException;
```
ravipesala: Even here also. same comment as above

core/src/main/java/org/carbondata/query/carbon/processor/impl/AggregateQueryBlockProcessor.java
```diff
@@ -43,12 +44,22 @@ public AggregateQueryBlockProcessor(BlockExecutionInfo tableBlockExecutionInfos,
...
-      this.scannerResultAggregator.aggregateData(blockletScanner.scanBlocklet(blocksChunkHolder));
+      try {
+        {
+          blocksChunkHolder.setDataBlock(dataBlockIterator.next());
+          blocksChunkHolder.reset();
+          this.scannerResultAggregator
+              .aggregateData(blockletScanner.scanBlocklet(blocksChunkHolder));
+        }
+      } catch (Exception e) {
+        // TODO Auto-generated catch block
```
ravipesala: remove the todo

sujith71955:
Handled review comments from ravindra.

core/src/main/java/org/carbondata/core/iterator/CarbonIterator.java
```diff
@@ -19,20 +19,15 @@
...
-
 /**
  * CarbonIterator adds default implement for remove. This is required for Java 7.
+ *
  * @param <E>
  */
-public abstract class CarbonIterator<E> implements Iterator<E> {
-
-  @Override public abstract boolean hasNext();
+public abstract class CarbonIterator<E> {
```
ravipesala: Any reason why `Iterator` implements is removed?

core/src/main/java/org/carbondata/query/carbon/executor/internal/impl/QueryRunner.java
```diff
@@ -66,7 +72,11 @@ public QueryRunner(BlockExecutionInfo executionInfo) {
     try {
       this.dataBlockProcessor.processBlock();
     } catch (Exception e) {
-      e.printStackTrace();
+      if (e.getCause() instanceof FilterUnsupportedException) {
+        throw new Exception(e.getCause());
```
ravipesala: I guess it supposed to be `throw new RuntimeException(e)`

core/src/main/java/org/carbondata/query/carbon/result/iterator/ChunkRawRowIterartor.java
```diff
@@ -28,8 +30,9 @@ public ChunkRawRowIterartor(CarbonIterator<BatchRawResult> iterator) {
    * throwing an exception.)
    *
    * @return {@code true} if the iteration has more elements
+   * @throws QueryExecutionException
    */
-  @Override public boolean hasNext() {
+  @Override public boolean hasNext()  {
```
ravipesala: please don't add unnecessary changes.

core/src/main/java/org/carbondata/query/carbon/executor/internal/impl/QueryRunner.java
```diff
@@ -66,7 +72,11 @@ public QueryRunner(BlockExecutionInfo executionInfo) {
...
+      } else {
+        LOGGER.error(e);
+      }
```
ravipesala: Just log the error and throw back same like `throw e`

core/src/main/java/org/carbondata/query/carbon/result/iterator/DetailQueryResultIterator.java
```diff
@@ -65,7 +64,7 @@ public DetailQueryResultIterator(List<BlockExecutionInfo> infos,
     try {
       result = executor.executeQuery(blockExecutionInfos, blockIndexToBeExecuted);
     } catch (QueryExecutionException e) {
-      LOGGER.error(e, e.getMessage());
+      throw new RuntimeException(e.getCause());
```
ravipesala: it should be `throw new RuntimeException(e)` . So that we can get stack trace on caller

core/src/main/java/org/carbondata/query/carbon/result/iterator/DetailQueryResultIterator.java
```diff
@@ -81,6 +80,7 @@ public DetailQueryResultIterator(List<BlockExecutionInfo> infos,
       } else {
         return new BatchResult();
       }
+
```
ravipesala: don't add space

core/src/main/java/org/carbondata/query/carbon/result/iterator/RawResultIterator.java
```diff
@@ -82,30 +83,28 @@ public RawResultIterator(CarbonIterator<BatchRawResult> detailRawQueryResultIter
...
   /**
    * for fetching the row with out incrementing counter.
+   *
    * @return
+   * @throws QueryExecutionException
    */
-  public Object[] fetch(){
-    if(hasNext())
-    {
+  public Object[] fetch() throws QueryExecutionException {
```
ravipesala: is throws required here?

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
```diff
@@ -326,37 +326,25 @@ public Boolean getBoolean() throws FilterUnsupportedException {
     if (null != expressionResults) {
       result = prime * result + expressionResults.hashCode();
     } else if (null != value) {
-      result = prime * result + value.hashCode();
+      result = prime * result + value.toString().hashCode();
```
ravipesala: any reason added to toString? better remove it

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
```diff
@@ -326,37 +326,25 @@ public Boolean getBoolean() throws FilterUnsupportedException {
...
-  public boolean equals(Object obj) {
-    if (null == obj) {
-      return false;
-    }
-
+  @Override public boolean equals(Object obj) {
     if (!(obj instanceof ExpressionResult)) {
       return false;
     }
     // CHECKSTYLE:ON
```
ravipesala: Please remove `CHECKSTYLE:ON`

core/src/main/java/org/carbondata/query/expression/conditional/EqualToExpression.java
```diff
@@ -77,7 +77,8 @@ public EqualToExpression(Expression left, Expression right) {
         result = val1.getDecimal().compareTo(val2.getDecimal()) == 0;
         break;
       default:
-        break;
+        throw new FilterUnsupportedException(
+            "DataType: " + val1.getDataType() + " not supported for the filter expression");
```
ravipesala: Better add class name in message to let user know about what expression it is.

core/src/main/java/org/carbondata/query/filters/FilterExpressionProcessor.java
```diff
@@ -62,9 +63,10 @@
    * @param tableIdentifier ,contains carbon store informations
    * @return a filter resolver tree
    * @throws QueryExecutionException
+   * @throws FilterUnsupportedException
    */
   public FilterResolverIntf getFilterResolver(Expression expressionTree,
-      AbsoluteTableIdentifier tableIdentifier) throws QueryExecutionException {
+      AbsoluteTableIdentifier tableIdentifier) throws FilterUnsupportedException {
```
ravipesala: I guess it is proper to keep `QueryExecutionException` in all places. Because `FilterUnsupportedException` is very specific exception. The operations can cause any type of exception. So we can just wrap the the `FilterUnsupportedException` with `QueryExecutionException`. Better don't do many changes.

ravipesala:
@sujith71955 Please don't do many changes for throwing `FilterUnsupportedException`, we already throwing checked exception `QueryExecutionException` it is good to enough to be generic for all types of exceptions. Any other new exception like `FilterUnsupportedException` can be wrapped by `QueryExecutionException` and throw

core/src/main/java/org/carbondata/core/iterator/CarbonIterator.java
kumarvishal09: Change CarbonIterator to interface

core/src/main/java/org/carbondata/query/carbon/executor/internal/impl/QueryRunner.java
```diff
@@ -66,7 +72,11 @@ public QueryRunner(BlockExecutionInfo executionInfo) {
     try {
       this.dataBlockProcessor.processBlock();
     } catch (Exception e) {
-      e.printStackTrace();
+      if (e.getCause() instanceof FilterUnsupportedException) {
```
kumarvishal09: no need to check instance of FilterUnsupportedException, add one catch block for FilterUnsupportedException

core/src/main/java/org/carbondata/query/carbon/executor/internal/impl/InternalDetailQueryExecutor.java
```diff
@@ -92,22 +95,30 @@ public InternalDetailQueryExecutor() {
...
+        } catch (ExecutionException e) {
+
+          throw new QueryExecutionException(e.getCause());
+        }
+      }
       return scannedResultProcessor.getQueryResultIterator();
     } catch (QueryExecutionException exception) {
-      LOGGER.error(exception, exception.getMessage());
-      throw new QueryExecutionException(exception);
+      throw new QueryExecutionException(exception.getCause());
```
kumarvishal09: throw new QueryExecutionException(exception.getCause()); better do not change

core/src/main/java/org/carbondata/query/carbon/executor/internal/impl/QueryRunner.java
kumarvishal09: add one more catch block for FilterUnsupportedException no need add instance check

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
sujith71955: There is an DTS issue and cause was in hashcode implementation of ExpressionResult the same two  filter members comparison was failing , the reference check because of diff hashcode values and getting failed in case of same filter member comparison present in 'IN' filter. because of this equals() not getting called where the filter member string value is been matched

core/src/main/java/org/carbondata/query/filters/FilterExpressionProcessor.java
sujith71955: This exception categorizes the query since the query is a filter type query and the FilterUnsupportedException  is been thrown only from filter layer, it directly conveys that the issue is present in the existing filter conditions

sujith71955:
retest this please

core/src/main/java/org/carbondata/query/expression/conditional/EqualToExpression.java
sujith71955: Actually by failing we are just  conveying user that the particular data type is not supported which is been used in the particular filter, so is it really required to provide the expression, and also more effort is required.

kumarvishal09:
retest this please

PR #624, opened 2016-06-10
[Issue 618]Supported Spark 1.6 in Carbondata

ravipesala:
Spark 1.6 changes approach of aggregation and changed the interfaces.  So all old implementations need to be either rewritten or deleted in this PR.

MeasureAggregator object is no longer supported in Spark 1.6 as UDT. As Spark 1.6 is completely based on UnSafeRow and codegen. So it try to read only the sqlType of UDT.
So approach is changed to avoid usage of  MeasureAggregator objects in Spark side aggregation.

Distinct aggregation like count-distinct or sum-distinct is no longer supported to push down to spark layer. Spark changes the plan completely to support distinct count in its way, so carbon cannot get a chance to push down distinct aggregation.

Interface identification and cleanup would be done another PR


ravipesala:
I am not sure, to which branch this PR can be merged? Should be merged to master or new branch or westlake? @jackylk Please help me.

jackylk:
@ravipesala I think carbon-spark integration module should provide spark 1.6 integration support as it can be directly used by community users who is running on spark 1.6. So I incline to merge it to master.

sraghunandan:
This would be possible if u have separate modules /packages for 1.5.2 and
1.6
We should be able to compile using Pom profiles
Else merging to master would fail 1.5.2

Vimal-Das:
@ravipesala , can we have support for both 1.5 and 1.6 as separate build options similar to how Spark supports different versions of Hadoop.

ravipesala:
@Vimal-Das There is a difference in Spark supporting different versions of Hadoop comparing to carbon supporting different versions on Spark.
Hadoop client library is standard and interfaces won't change and they make sure the backward compatability with old client libraries. That's why spark can just change the hadoop jar version while building would be enough.
But when it comes to Carbon, we are not just using the standard spark client interfaces it has very deep integration to the spark sql layer. 
I guess we may have following two solutions.
1. Make the current code to branch as spark 1.5 and push the spark 1.6 to the master. So if user wants to use spark 1.5 then they need to take from branch.
Cons : Defect fixes need to keep sync between both branches 
2. The changed classes/interfaces need to be maintained in separate module or just a package inside integration/spark module. Pick the right version package while building using maven. But extracting the changed classes or interfaces  is not so simpler or cleaner.
Cons : Version to version interfaces changes are maintained different packages are not a good way. And also for spark 2.0 it would be a major change and not possible to do in this way.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonCatalystOperators.scala
```diff
@@ -94,7 +94,7 @@ case class ShowCubeCommand(schemaNameOp: Option[String]) extends LogicalPlan wit
   override def children: Seq[LogicalPlan] = Seq.empty
 
   override def output: Seq[Attribute] = {
-    Seq(AttributeReference("cubeName", StringType, nullable = false)(),
+    Seq(AttributeReference("tableName", StringType, nullable = false)(),
```
jackylk: So this PR is not just adapt to spark 1.6, it also change cubeName and schemaName, right? 
Did you change in all places?

integration/spark/src/main/scala/org/apache/spark/sql/agg/CarbonAggregationExpression.scala
```diff
@@ -0,0 +1,50 @@
...
+    }
+    val right = input2 match {
+      case g: ArrayData => g.getDouble(index)
+      case others => others.toString.toDouble
+    }
+    if (index == 1) {
+      (left + right).toLong
+    } else {
+      dataType match {
+        case d: DecimalType =>
```
jackylk: DecimalType is not supported?

integration/spark/src/main/scala/org/apache/spark/sql/agg/CarbonAverage.scala
```diff
@@ -0,0 +1,89 @@
...
+
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
+import org.apache.spark.sql.catalyst.dsl.expressions._
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.expressions.aggregate.DeclarativeAggregate
+import org.apache.spark.sql.types._
+
+@ExpressionDescription(
+  usage = "_FUNC_(x) - Returns the mean calculated from values of a group.")
+case class CarbonAverage(child: Expression) extends DeclarativeAggregate {
```
jackylk: Can we extend it from spark Average? I see only checkInputDataTypes is different?

integration/spark/src/main/scala/org/apache/spark/sql/agg/CarbonCount.scala
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.spark.sql.catalyst.expressions.aggregate.DeclarativeAggregate
+import org.apache.spark.sql.types._
+
+// scalastyle:off line.size.limit
+@ExpressionDescription(
+  usage = """_FUNC_(*) - Returns the total number of retrieved rows, including rows containing NULL values.
+    _FUNC_(expr) - Returns the number of rows for which the supplied expression is non-NULL.
+    _FUNC_(DISTINCT expr[, expr...]) - Returns the number of rows for which the supplied expression(s) are unique and non-NULL.""")
+// scalastyle:on line.size.limit
+case class CarbonCount(
```
jackylk: Can we extend it from spark Count?

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonStrategies.scala
```diff
@@ -188,59 +182,138 @@ class CarbonStrategies(sqlContext: SQLContext) extends QueryPlanner[SparkPlan] {
...
-                groupingExpressions,
-                partialComputation,
-                s)(sqlContext))(sqlContext) :: Nil
+            aggregatePlan(groupingExpressions, namedGroupingAttributes, s)
+        }
+      }
+    }
+
+    // TODO: It is duplicated code from spark. Need to find a way
+    private def aggregatePlan(groupingExpressions: Seq[Expression],
```
jackylk: This is only needed for pushdown path, right?

integration/spark/src/main/scala/org/apache/spark/sql/CarbonCatalystOperators.scala
ravipesala: Yes, in spark 1.6 they have introduced TableIdentifier so along with those changes I have updated cubeName and schemaName

integration/spark/src/main/scala/org/apache/spark/sql/agg/CarbonAggregationExpression.scala
ravipesala: Yes, It is not yet supported. I will support it.

integration/spark/src/main/scala/org/apache/spark/sql/agg/CarbonAverage.scala
ravipesala: Mainly `updateExpressions` is changed. I will try to extend if possible

integration/spark/src/main/scala/org/apache/spark/sql/agg/CarbonCount.scala
ravipesala: No it is not possible as Spark Count is case class

integration/spark/src/main/scala/org/apache/spark/sql/hive/CarbonStrategies.scala
ravipesala: yes , It is only needed for push down path, for push up we use `CarbonRawStrategies`

PR #627, opened 2016-06-11
[Issue - 626] csv file having some rows not in proper format is not treating them to the bad records.

### LINKED ISSUE #626: csv file having some rows not in proper format is not pushing them to the bad records.
csv file having some rows not in proper format is not pushing them to the bad records.
for example:
if csv data is as given below
state,country,population 
UP,INDIA,23498900
Delhi
Mumbai,INDIA,383289

then while loading the data, the second row should be treated as bad record as it is not in proper csv format.

mohammadshahidkhan:
fixed  #626

### DIFF
--- a/processing/src/main/java/org/carbondata/processing/csvreaderstep/BlockDataHandler.java
+++ b/processing/src/main/java/org/carbondata/processing/csvreaderstep/BlockDataHandler.java
@@ -27,6 +27,8 @@ import java.util.List;
 import org.carbondata.core.constants.CarbonCommonConstants;
 import org.carbondata.core.datastorage.store.impl.FileFactory;
 import org.carbondata.core.load.BlockDetails;
+import org.carbondata.processing.surrogatekeysgenerator.csvbased.BadRecordslogger;
+import org.carbondata.processing.util.CarbonDataProcessorUtil;
 
 import org.apache.commons.vfs.FileObject;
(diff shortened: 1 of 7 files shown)


processing/src/main/java/org/carbondata/processing/csvreaderstep/BlockDataHandler.java
```diff
@@ -584,7 +582,11 @@ else if (data.enclosure != null && data.enclosureMatcher
             + " conversion errors on line ", conversionExceptions, exceptionFields, outputRowData);
       }
 
-      return outputRowData;
+      if (outputIndex > 0 && outputIndex < meta.getInputFields().length) {
+        return new BlockDataWrapper(true, outputRowData);
```
Vimal-Das: Instead of creating a wrapper object, call the bad record logger from here itself.
1) you will save memory by not holding bad records in memory
2) reduce the number of if checks in CsvInput.

Do not use recursion, instead use simple while loop in the readOneRow method.

mohammadshahidkhan: Fixed review comment

Vimal-Das:
@mohammadshahidkhan Check -4 tests

Vimal-Das:
retest this please

Vimal-Das:
LGTM, Merged to master.

PR #628, opened 2016-06-11
[Bug]Null measure value issue Data mismatch

kumarvishal09:
Fixed Null Measure Value issue , it was showing unique value need to show null in case of detail query
in case of aggregate query need to show based on aggregator
sum , avg, max, min, sum distinct will show null
for count and distinct count 0


sujith71955:
LGTM

Issue #635, opened 2016-06-13
Select Avg(StringType)+IntType cause stage failure


Jay357089:
when execute "Select Avg(StringType)+IntType from table", it will cause unresolved exception and stage failure. i will raise a pr to handle this.

gvramana:
What is the behaviour in case of hive

Jay357089:
@gvramana hive shows "null".

Jay357089:
handled. close it.

PR #636, opened 2016-06-13
[issue-635]add handle for avg(stringType)+IntType query

Jay357089:
issue #635 

the stage failure is because avg(StringType) is resolved as String Type and IntType resolved as DoubleType, the type of both plus can not be resolved. 
i cast it to DoubleType.


Jay357089:
@kumarvishal09  @QiangCai  pls review.

kumarvishal09:
please handle for other aggregate function sum, max, min, sum distinct for example sum(String)+10 in this can also we need to show null

Jay357089:
@kumarvishal09 handled.

Jay357089:
@kumarvishal09 pls review and check in.

Jay357089:
retest this please

kumarvishal09:
retest this please

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -111,7 +111,8 @@ case class AverageCarbon(child: Expression, castedDataType: DataType = null)
     val partialSum = Alias(AverageCarbon(child), "PartialAverage")()
     SplitEvaluation(
       AverageCarbonFinal(partialSum.toAttribute,
-        if (child.dataType == IntegerType) DoubleType else child.dataType),
+        if (child.dataType == IntegerType || child.dataType == StringType) DoubleType
```
ravipesala: what if the datatype is timestamp?It fails in this case so please handle that scenario as well.

kumarvishal09:
@Jay357089 you can check how other aggregate functions are working you can use the same code

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
Jay357089: @ravipesala  i will handle

Jay357089:
@kumarvishal09 there is some difference between avg and other aggregation, because avg(int) should be double, and for dimension, int is handled as int, but for measure, int is handled as double. so, some specific handle should be done to avg.

Jay357089:
@ravipesala handled, also added test case. pls review and check in.

ravipesala:
LGTM

PR #642, opened 2016-06-13
Zookeeper Instance need to be created in the carbon env.

ravikiran23:
Creating the zookeeper client on the start up.
creating the singleton instance of zookeeperInit.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonEnv.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonEnv.scala
@@ -19,10 +19,6 @@ package org.apache.spark.sql
 
 import org.apache.spark.sql.hive.{CarbonMetastoreCatalog, HiveContext}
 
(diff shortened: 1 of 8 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/CarbonEnv.scala
```diff
@@ -33,6 +37,16 @@ object CarbonEnv {
       carbonEnv =
         CarbonEnv(sqlContext.asInstanceOf[CarbonContext],
           sqlContext.asInstanceOf[CarbonContext].catalog)
+
+      // creating zookeeper instance once.
+      // if zookeeper is configured as carbon lock type.
+      if (CarbonProperties.getInstance()
```
ravipesala: why don't you move this code to `ZookeeperInit.getInstance()` class. Any way it is singleton class right. 
Because adding code in CarbonEnv is not recommendable as it is to be deleted class, it is doing nothing.

ravipesala:
LGTM

PR #645, opened 2016-06-14
Fixed bug  select function like trim or length along with avg aggregator

ravipesala:
Following queries was not working 
select SID,length( MSISDN),max(LAYER1ID),min(LAYER1ID),avg(LAYER1ID) from smart_500 group by SID,length( MSISDN) order by SID limit 10;
select age,trim(imei),avg(gamepointid),avg(age) from table03 group by age,trim(imei);

It is because we are converting to carbon aggregates in `CarbonCatalystOperators` but going to detail query in `carbonOperators`, so it causes unexpected results. This PR add validation to made it sync between two classes


ravipesala:
@kumarvishal09 I don't have actual dataset to verify so please verify it with actual dataset and merge it.

kumarvishal09:
ok @ravipesala  I will check with actual data. Thank you

kumarvishal09:
retest this please

PR #647, opened 2016-06-14
[bug]add escape character option to global dictionary

eason-lyx:
while using spark-csv to load the data,by default,the escape char is "\".
so when the data in quote with "\",then spark-csv will skip \,like "ab\c"->"abc".
and in csv reader ,it will no skip.

so there is inconsistency in read in phase 1 - dictionary generation and phase 2- data loading.

then we provide an option to user to specify escape character.
@Vimal-Das please review.thanks.


QiangCai:
LGTM

eason-lyx:
retest this please

integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataload/TestLoadDataWithHiveSyntax.scala
```diff
@@ -201,10 +201,29 @@ class TestLoadDataWithHiveSyntax extends QueryTest with BeforeAndAfterAll {
...
+    sql("""
+           CREATE TABLE IF NOT EXISTS t3
+           (ID Int, date Timestamp, country String,
+           name String, phonetype String, serialname String, salary Int)
+           STORED BY 'org.apache.carbondata.format'
+           """)
+
+    sql(s"""
+           LOAD DATA LOCAL INPATH './src/test/resources/datawithbackslash.csv' into table t3
+           OPTIONS('ESCAPECHAR'='@')
```
QiangCai: why use '@' ,but not  '"'

QiangCai:
OK,  merge to master.

PR #652, opened 2016-06-15
[Bug] Failure in case of concurrent data loading of multiple tables

manishgupta88:
**Issue:** During concurrent data load if multiple tasks for same table were getting submitted to the same executor, then the templocation key variable set in carbonproperties instance was not unique because of which sometimes path mismatch was happening leading to load failure.

**Fix:** Made the tempLocationKey variable Unique


kumarvishal09:
retest this please

kumarvishal09:
@ravikiran23 in-progress extension is removed from temp segment folder name, Please handle this in compaction

PR #653, opened 2016-06-15
[Bug] NOT NULL should not display any null values in the output

sujith71955:
Problem:
NOT NULL should not display any null values in the output

 No_Dictionary columns wont support double/decimal type, but in case of integer type data this scenario has to be handled.
issue is happening since no dictionary integer type column block contain String value, so when filter comparison happens the actual data which is a non null value  present in block
will be compared with default null member and the null will not get filtered out.

Solution:
To solve this issue while performing data load the no dictionary integer/other type of column excluding String type contains any un parseable data
the system has to convert it to default null member value.

### DIFF
--- a/processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
+++ b/processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
@@ -72,6 +72,7 @@ import org.carbondata.processing.util.CarbonDataProcessorUtil;
 import org.carbondata.processing.util.RemoveDictionaryUtil;
 
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
+
 import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.row.RowMetaInterface;
 import org.pentaho.di.core.row.ValueMeta;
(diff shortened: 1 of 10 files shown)


ravipesala:
@sujith71955 , there are test cases are failing please fix them. And also add test case for this scenario.

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1750,7 +1750,14 @@ private void clearDictionaryCache() {
     }
   }
 
-  private void processnoDictionaryDim(int index, String dimensionValue, ByteBuffer[] out) {
+  private void processnoDictionaryDim(int index, String dimensionValue, String dataType,
+      ByteBuffer[] out) {
+    if (!(DataTypeUtil.getDataType(dataType) == DataType.STRING)) {
```
ravipesala: Don't convert to datatype for each row. Better convert them earlier.

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -910,14 +920,14 @@ private String getCarbonLocalBaseStoreLocation() {
       String columnName = metaColumnNames[j];
       String foreignKeyColumnName = foreignKeyMappingColumns[j];
       r[j] = ((String) r[j]).trim();
-      // TODO check if it is ignore dictionary dimension or not . if yes directly write byte buffer
-
+      // check if it is ignore dictionary dimension or not . if yes directly write byte buffer
       if (isNoDictionaryColumn[j]) {
-        processnoDictionaryDim(noDictionaryAndComplexIndexMapping[j],
-            (String) r[j], byteBufferArr);
+        String dataType = meta.dimColDataTypes.get(columnName.toLowerCase());
```
kumarvishal09: no need to get the data type from map as it will always check equals and in case of more number columns it will add a overhead, you can create a data type array and get the data type based on index

PR #655, opened 2016-06-15
[wip] test

Zhangshunyu:


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -1125,7 +1115,7 @@ class CarbonSqlParser()
   protected lazy val dimCol: Parser[Field] = anyFieldDef
 
   protected lazy val primitiveTypes =
-    STRING | INTEGER | TIMESTAMP | NUMERIC | BIGINT | DECIMAL | INT | DOUBLE
+    "string" | "integer" | "timestamp" | "numeric" | "bigint" | "decimal" | "int" | "double"
```
nareshpr: Can you try to prepare another keyword list which are allowed for tablename and dbname and use it while parsing?

Zhangshunyu:
retest this please

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
Zhangshunyu: @nareshpr OK，I will do it

PR #657, opened 2016-06-15
[issue-656] fix load data when int column contains integer.min_value

### LINKED ISSUE #656: Query result wrong when int column contains max value and min value of integer
create table:
`CREATE TABLE ant(imei string,age int) STORED BY 'org.apache.carbondata.format'`

load data
`LOAD DATA INPATH 'D:/hadoop/srcdata/ayush1.csv' INTO table t3 options  ('DELIMITER'=',', 'QUOTECHAR'='\"', 'FILEHEADER'= 'imei,age'`

test data like: 
```
imei,age
1AA1,10
1AA2,26
1AA3,10
1AA4,10
1AA5,20
1AA6,10
1AA7,10
1AA8,10
1AA9,10
1AA10,10
1AA11,-2147483648
1AA12,2147483647
```
query result is not correct

foryou2030:
fix #656


foryou2030:
@kumarvishal09 
pls review it

core/src/main/java/org/carbondata/core/util/ValueCompressionUtil.java
```diff
@@ -584,11 +593,12 @@ private static Object compressNonDecimalMaxMin(DataType changedDataType, double[
    */
   public static ValueCompressonHolder.UnCompressValue unCompressMaxMin(DataType compDataType,
       DataType actualDataType) {
-    if (actualDataType == DataType.DATA_LONG) {
+    if (actualDataType == DataType.DATA_LONG || actualDataType == DataType.DATA_BIGINT) {
```
kumarvishal09: why this condition is required? this can be called for other data types ..may be our test case does not have this max min cases but this condition is wrong

foryou2030: the compressType for  DATA_BIGINT only use COMPRESSION_TYPE.NONE, 
So i think that this condition has no impact to current flow, it is not required. Keep or remove it both is ok, yes?

foryou2030:
@kumarvishal09 i have removed some no used condition(DATA_BIGINT), pls review. thanks

kumarvishal09:
retest this please

kumarvishal09:
/**
   * uncompress data 1. value = maxValue - subValue 2. value: int->double
   */
  public static ValueCompressonHolder.UnCompressValue unCompressMaxMin(DataType compDataType,
      DataType actualDataType) {
    if (actualDataType == DataType.DATA_LONG) {
      switch (compDataType) {
        case DATA_BYTE:
          return new UnCompressMaxMinByteForLong();
        case DATA_LONG:
          return new UnCompressMaxMinDefaultLong();
        default:
          return new UnCompressMaxMinDefaultLong();
      }
    } else {

I think this if condition is not correct please remove line no:592

foryou2030:
@kumarvishal09 yes, i have disscued with @Jay357089 ,this condition is no used currently. I removed it.

kumarvishal09:
retest this please

Issue #660, opened 2016-06-15
All Queries with case insensitive table name should not fail.


ManoharVanam:
Below queries are failing If table name is case insensitive
1. Show Segments
2. Drop table
3. Delete Segments

ManoharVanam:
Fixed all scenarios.

Issue #648, opened 2016-06-14
UT failed in CarbonTestSuite with native code


Hexiaoqiao:
when i run UT in my localhost, it failed since 2 exceptions throws in {DataCompactionCardinalityBoundryTest} and {DataCompactionNoDictionaryTest}.

my local env:
OS:Darwin Kernel Version 15.5.0,RELEASE_X86_64
JDK:jdk1.7.0_75
Scala:2.11.5

exception trace:
> DataCompactionNoDictionaryTest:
Exception encountered when invoking run on a nested suite - alter table failed. table not found: default_noDictionaryCompaction *** ABORTED ***
  java.lang.RuntimeException: alter table failed. table not found: default_noDictionaryCompaction
  at scala.sys.package$.error(package.scala:27)
  at org.apache.spark.sql.execution.command.AlterTableCompaction.run(carbonTableSchema.scala:1162)
  at org.apache.spark.sql.execution.ExecutedCommand.sideEffectResult$lzycompute(commands.scala:57)
  at org.apache.spark.sql.execution.ExecutedCommand.sideEffectResult(commands.scala:57)
  at org.apache.spark.sql.execution.ExecutedCommand.doExecute(commands.scala:69)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$5.apply(SparkPlan.scala:140)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$5.apply(SparkPlan.scala:138)
  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:147)
  at org.apache.spark.sql.execution.SparkPlan.execute(SparkPlan.scala:138)
  at org.apache.spark.sql.SQLContext$QueryExecution.toRdd$lzycompute(SQLContext.scala:933)
  ...

> DataCompactionCardinalityBoundryTest:
Exception encountered when invoking run on a nested suite - alter table failed. table not found: default_cardinalityTest *** ABORTED ***
  java.lang.RuntimeException: alter table failed. table not found: default_cardinalityTest
  at scala.sys.package$.error(package.scala:27)
  at org.apache.spark.sql.execution.command.AlterTableCompaction.run(carbonTableSchema.scala:1162)
  at org.apache.spark.sql.execution.ExecutedCommand.sideEffectResult$lzycompute(commands.scala:57)
  at org.apache.spark.sql.execution.ExecutedCommand.sideEffectResult(commands.scala:57)
[...]

PR #662, opened 2016-06-15
[Issue-660] Show segments query should not fail, if table name is case insensitive

ManoharVanam:
Refer : #660 
Show segments query should not fail, if table name is case insensitive

### DIFF
--- a/integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataretention/DataRetentionTestCase.scala
+++ b/integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataretention/DataRetentionTestCase.scala
@@ -96,7 +96,7 @@ class DataRetentionTestCase extends QueryTest with BeforeAndAfterAll {
 
   test("RetentionTest_withoutDelete") {
     checkAnswer(
-      sql("SELECT country, count(salary) AS amount FROM DataRetentionTable WHERE country" +
+      sql("SELECT country, count(salary) AS amount FROM dataretentionTable WHERE country" +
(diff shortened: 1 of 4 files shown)


integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataretention/DataRetentionTestCase.scala
```diff
@@ -96,7 +96,7 @@ class DataRetentionTestCase extends QueryTest with BeforeAndAfterAll {
 
   test("RetentionTest_withoutDelete") {
     checkAnswer(
-      sql("SELECT country, count(salary) AS amount FROM DataRetentionTable WHERE country" +
+      sql("SELECT country, count(salary) AS amount FROM dataretentionTable WHERE country" +
```
ravipesala: why you have changed to lower case in test case?Even if it is mixed case also supposed to work right?

ravipesala:
Also the test cases inside `DataCompactionCardinalityBoundryTest` not running, it is same case sensitive issue. Check the issue https://github.com/HuaweiBigData/carbondata/issues/648 . Please make sure that also fixed in this PR.

integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataretention/DataRetentionTestCase.scala
ManoharVanam: Here , Iam checking mixed case : dataretentionTable

ManoharVanam:
retest this please

ravipesala:
LGTM

PR #663, opened 2016-06-15
Bug fix in the minor compaction segment size identification.

ravikiran23:
Handling of the locking mechanism for compaction using wrapper thread.
improving the logs for compaction.

### DIFF
--- a/integration/spark/src/main/java/org/carbondata/integration/spark/merger/RowResultMerger.java
+++ b/integration/spark/src/main/java/org/carbondata/integration/spark/merger/RowResultMerger.java
@@ -215,6 +215,7 @@ public class RowResultMerger {
     carbonFactDataHandlerModel.setDatabaseName(schemaName);
     carbonFactDataHandlerModel.setTableName(tableName);
     carbonFactDataHandlerModel.setMeasureCount(segprop.getMeasures().size());
+    carbonFactDataHandlerModel.setCompactionFlow(true);
     carbonFactDataHandlerModel
         .setMdKeyLength(segprop.getDimensionKeyGenerator().getKeySizeInBytes());
     carbonFactDataHandlerModel.setStoreLocation(tempStoreLocation);
--- a/integration/spark/src/main/java/org/carbondata/integration/spark/merger/RowResultMerger.java
+++ b/integration/spark/src/main/java/org/carbondata/integration/spark/merger/RowResultMerger.java
@@ -127,9 +127,8 @@ public class RowResultMerger {
   /**
    * Merge function
    *
-   * @throws SliceMergerException
    */
-  public boolean mergerSlice() throws SliceMergerException {
+  public boolean mergerSlice() {
     boolean mergeStatus = false;
     int index = 0;
     try {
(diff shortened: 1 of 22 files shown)


integration/spark/src/main/java/org/carbondata/spark/merger/CarbonDataMergerUtil.java
```diff
@@ -409,13 +407,6 @@ private static boolean isTwoDatesPresentInRequiredRange(Date segDate1, Date segD
       // in case of major compaction the size doesnt matter. all the segments will be merged.
       if (totalLength < (compactionSize * 1024 * 1024)) {
```
Vimal-Das: We should externalize this size limits to carbon.properties

integration/spark/src/main/java/org/carbondata/integration/spark/merger/RowResultMerger.java
```diff
@@ -215,6 +215,7 @@ private CarbonFactDataHandlerModel getCarbonFactDataHandlerModel(CarbonLoadModel
     carbonFactDataHandlerModel.setDatabaseName(schemaName);
     carbonFactDataHandlerModel.setTableName(tableName);
     carbonFactDataHandlerModel.setMeasureCount(segprop.getMeasures().size());
+    carbonFactDataHandlerModel.setCompactionFlow(true);
```
gvramana: write a comment why this is required

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -70,7 +70,7 @@ class CarbonMergerRDD[K, V](
       val carbonSparkPartition = theSplit.asInstanceOf[CarbonSparkPartition]
 
       val tempLocationKey: String = carbonLoadModel.getDatabaseName + '_' + carbonLoadModel
-        .getTableName
+        .getTableName + carbonLoadModel.getTaskNo
```
gvramana: You have to use temp location, to pass, cannot set it in CarbonProperties

processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -342,13 +342,25 @@ private void initParameters(CarbonFactDataHandlerModel carbonFactDataHandlerMode
...
-    } catch (NumberFormatException exc) {
-      numberOfCores = Integer.parseInt(CarbonCommonConstants.NUM_CORES_DEFAULT_VAL);
+
+    if (carbonFactDataHandlerModel.isCompactionFlow()) {
+      try {
+        numberOfCores = Integer.parseInt(CarbonProperties.getInstance()
+            .getProperty(CarbonCommonConstants.NUM_CORES_COMPACTING,
+                CarbonCommonConstants.NUM_CORES_DEFAULT_VAL));
+      } catch (NumberFormatException exc) {
+        numberOfCores = Integer.parseInt(CarbonCommonConstants.NUM_CORES_DEFAULT_VAL);
```
gvramana: write logger, falling back to default

integration/spark/src/main/java/org/carbondata/integration/spark/merger/RowResultMerger.java
ravikiran23: added comments.

processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
ravikiran23: writing the logger.

integration/spark/src/main/java/org/carbondata/spark/merger/CarbonDataMergerUtil.java
ravikiran23: compactionSize is taken from the carbon properties only.

Issue #664, opened 2016-06-15
select count(joinDate) from table_x is failing direct dictionary column


mohammadshahidkhan:
select count(joinDate) from table_x is failing direct dictionary column

mohammadshahidkhan:
org.carbondata.query.carbon.executor.exception.QueryExecutionException: 
	at org.carbondata.query.carbon.processor.impl.AggregateQueryBlockProcessor.processBlock(AggregateQueryBlockProcessor.java:57)
	at org.carbondata.query.carbon.executor.internal.impl.QueryRunner.call(QueryRunner.java:72)
	at org.carbondata.query.carbon.executor.internal.impl.QueryRunner.call(QueryRunner.java:36)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
	at org.carbondata.query.carbon.aggregator.dimension.impl.FixedLengthDimensionAggregator.aggregateDimensionData(FixedLengthDimensionAggregator.java:102)
	at org.carbondata.query.carbon.aggregator.DataAggregator.aggregateData(DataAggregator.java:71)
	at org.carbondata.query.carbon.aggregator.impl.MapBasedResultAggregator.aggregateData(MapBasedResultAggregator.java:109)
	at org.carbondata.query.carbon.processor.impl.AggregateQueryBlockProcessor.processBlock(AggregateQueryBlockProcessor.java:55)
	... 6 more

mohammadshahidkhan:
Fixed with PR #665

PR #665, opened 2016-06-15
[issue- 664] select count(joinDate) from table_x is failing for direct dictionary column

mohammadshahidkhan:


mohammadshahidkhan:
Fixed #664

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/FixedLengthDimensionAggregator.java
```diff
@@ -22,7 +22,10 @@
 
 import org.carbondata.core.cache.dictionary.Dictionary;
```
kumarvishal09: its better to handle dimension aggregation for direct dictionary dimension in different class instead of adding multiple responsibility in this class

mohammadshahidkhan: Fixed review comments

mohammadshahidkhan:
Please review and merge this its blocking for PR#615

mohammadshahidkhan:
Reduced test case is not failure previous run was having addition TestLoadDataWithMaxMinInteger 
test case

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Below method will be used to aggregate the dimension data
+   *
+   * @param scannedResult scanned result
+   * @param aggeragtor    aggregator used to aggregate the data
+   */
+  @Override public void aggregateDimensionData(AbstractScannedResult scannedResult,
+      MeasureAggregator[] aggeragtor) {
+    byte[] dimensionData = scannedResult.getDimensionKey(blockIndex);
+    int surrogateKey = CarbonUtil.getSurrogateKey(dimensionData, buffer);
```
kumarvishal09: i this code is not required here as you need to apply same formula the way you generated the surrogate key, this is only for dictionary column. you can direct dictionary key generator to get the surrogate

kumarvishal09: not valid i will merge

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
```diff
@@ -0,0 +1,121 @@
...
+    if (surrogateKey == 1) {
```
kumarvishal09: is this scenario scenario will come in case of direct dictionary? in data loading if parsing is failing for direct dictionary dimension are we returning one if not then return 1. Then invalid data treated as null

mohammadshahidkhan: no in direct dictionary we are returning -1,
Will fix this.

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
```diff
@@ -0,0 +1,117 @@
...
+   *
+   * @param scannedResult scanned result
+   * @param aggeragtor    aggregator used to aggregate the data
+   */
+  @Override public void aggregateDimensionData(AbstractScannedResult scannedResult,
+      MeasureAggregator[] aggeragtor) {
+    byte[] dimensionData = scannedResult.getDimensionKey(blockIndex);
+    int surrogateKey = CarbonUtil.getSurrogateKey(dimensionData, buffer);
+    Object dataBasedOnDataType = null;
+    dataBasedOnDataType =
```
ravipesala: combine 97 and 98 lines

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
```diff
@@ -0,0 +1,117 @@
...
+        (long) directDictionaryGenerator.getValueFromSurrogate(surrogateKey) / 1000;
+
+    // if data is null then no need to aggregate
+    if (null == dataBasedOnDataType) {
```
ravipesala: I don't see any case it could be null. Please check once while loading direct dictionary values what surrogateKey is added for null.

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
mohammadshahidkhan: fixed

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Class which will be used to aggregate the direct dictionary dimension data
+ */
+public class DirectDictionaryDimensionAggregator implements DimensionDataAggregator {
+
+  /**
+   * info object which store information about dimension to be aggregated
+   */
+  protected DimensionAggregatorInfo dimensionAggeragtorInfo;
```
ravipesala: make it private

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
mohammadshahidkhan: yes for direct dictionary it could not be null, will fix this

mohammadshahidkhan:
Fixed review comment please  review

core/src/main/java/org/carbondata/query/carbon/aggregator/dimension/impl/DirectDictionaryDimensionAggregator.java
```diff
@@ -0,0 +1,114 @@
...
+   * @param scannedResult scanned result
+   * @param aggeragtor    aggregator used to aggregate the data
+   */
+  @Override public void aggregateDimensionData(AbstractScannedResult scannedResult,
+      MeasureAggregator[] aggeragtor) {
+    byte[] dimensionData = scannedResult.getDimensionKey(blockIndex);
+    int surrogateKey = CarbonUtil.getSurrogateKey(dimensionData, buffer);
+    Object dataBasedOnDataType =
+        (long) directDictionaryGenerator.getValueFromSurrogate(surrogateKey) / 1000;
+
```
ravipesala: where did you handle the data null case?

ravipesala:
@mohammadshahidkhan why in direct dictionary it could not be null? while loading data if it comes null then are we adding it to bad records?

mohammadshahidkhan:
@Vimal-Das decided if the timestamp data is not parseable then move the same to bad record.

ravipesala:
ok LGTM

PR #666, opened 2016-06-16
[BUG]Improve first time query performance and block pruning issue 

kumarvishal09:
1.Added code to provide parallel segment block loading in driver.
2.Added code to provide parallel blocks loading in executor to improve first time query execution performance
3. In case of filter query if non of the blocks where .selected after applying max,min then also first block was getting added because of this more no of task was getting launched in executor and it was hitting the performance, fixed the same 
4. Added test case to load block and segment blocks parallely


core/src/main/java/org/carbondata/core/carbon/datastore/BlockIndexStore.java
```diff
@@ -157,16 +212,13 @@ public void removeTableBlocks(List<TableBlockInfo> removeTableBlocksInfos,
...
-    synchronized (lockObject) {
-      Map<TableBlockInfo, AbstractIndex> map = tableBlocksMap.get(absoluteTableIdentifier);
-      // if there is no loaded blocks then return
-      if (null == map) {
-        return;
-      }
-      for (TableBlockInfo blockInfos : removeTableBlocksInfos) {
-        map.remove(blockInfos);
-      }
+    Map<TableBlockInfo, AbstractIndex> map = tableBlocksMap.get(absoluteTableIdentifier);
```
sujith71955: provide proper name for the reference

core/src/main/java/org/carbondata/core/carbon/datastore/BlockIndexStore.java
```diff
@@ -93,54 +101,101 @@ public static BlockIndexStore getInstance() {
...
     }
     return loadedBlocksList;
   }
 
   /**
+   * Method to add table level lock if lock is not present for the table
+   *
+   * @param absoluteTableIdentifier
+   */
+  private void addTableLockObject(AbsoluteTableIdentifier absoluteTableIdentifier) {
```
sujith71955: i think we can make API itself synchronized?

core/src/main/java/org/carbondata/core/carbon/datastore/BlockIndexStore.java
kumarvishal09: ok

core/src/main/java/org/carbondata/core/carbon/datastore/BlockIndexStore.java
kumarvishal09: ok

sujith71955:
Done.

PR #667, opened 2016-06-16
[Bug] Sort index not created properly for null members in case of String type

sujith71955:
SortModel reference was not sorting the forward dictionary values when null member/nonparseable was present in the column data other than string.

In filter layer also when IS NOT NULL filter resolve happens the filter expression has to be evaluated with all the filter values when the column data has both null and unparseable data in different data type column other than String.


kumarvishal09:
retest this please

kumarvishal09:
retest this please

PR #668, opened 2016-06-16
[issue-581]fix the bug of load parse failure when tblname is in carbon's datatype keywords in DDL.

Zhangshunyu:
We discussed on this issue and had a conclusion:
Only data type ignore keywords, the others keep as reserved keywords of carbon.


Zhangshunyu:
@nareshpr @sujith71955 
pls review this pr, thank you!

integration/spark/src/test/scala/org/carbondata/integration/spark/testsuite/dataload/TestLoadTblNameIsKeyword.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  test("test load data whose name is a keyword of carbon ddl parser") {
+    sql("""
+          LOAD DATA INPATH './src/test/resources/tblNameWithDataTypeKeyword.csv' INTO TABLE
+          STRING OPTIONS('DELIMITER'=',', 'QUOTECHAR'= '"')
+        """)
+    checkAnswer(
+      sql("""
+            SELECT count(*) from STRING
```
sujith71955: Please add test case which can validate the case in-sensitiveness of data type while parsing

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -1347,7 +1348,8 @@ class CarbonSqlParser()
     }
 
   protected lazy val deleteLoadsByDate: Parser[LogicalPlan] =
-    DELETE ~> FROM ~> CUBE ~> (ident <~ ".").? ~ ident ~ (WHERE ~> (ident <~ BEFORE) ~ stringLit) <~
+    DELETE ~> FROM ~> CUBE ~> ((ident | primitiveTypes) <~ ".").? ~ (ident | primitiveTypes) ~
+      (WHERE ~> (ident <~ BEFORE) ~ stringLit) <~
```
sujith71955: i think in where also we need to consider primitiveTypes, Please relook into this

integration/spark/src/test/resources/tblNameWithDataTypeKeyword.csv
```diff
@@ -0,0 +1,21 @@
+imei,deviceInformationId,AMSize,channelsId,ActiveCountry,Activecity,gamePointId,productionDate,deliveryDate,deliverycharge
```
sujith71955: why this csv is required, can we use existing csv?

integration/spark/src/test/scala/org/carbondata/integration/spark/testsuite/dataload/TestLoadTblNameIsKeyword.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.spark.sql.common.util.CarbonHiveContext._
+import org.apache.spark.sql.common.util.QueryTest
+import org.scalatest.BeforeAndAfterAll
+import org.apache.spark.sql.Row
+
+/**
+ * Test Class for data loading into table whose name is key word
+ *
+ */
+class TestLoadTblNameIsKeyword extends QueryTest with BeforeAndAfterAll {
```
sujith71955: Better to add more testcases inorder to verify the above mentioned behavior in all ddls supported by carbon, its also has impact in other ddls

Zhangshunyu:
@sujith71955 pls check.

Zhangshunyu:
@gvramana data type is not keywords of spark...
spark has many parsers，ddlparser、sparksqlparser、hiveparser，we only have carbonsqlparser...we only care about use only one parser to parser it，I discussed with naresh，we think this is the bestway，we only modify little code.

Zhangshunyu:
if u want do as spark，then the datatype will move from keywords using regex to parse，I think this way is not a better way.

QiangCai:
@gvramana I agree with you to keep consistent with spark.

Zhangshunyu:
please look at this pr #704

PR #669, opened 2016-06-16
Optimize HDFSCarbonFile.listFiles(CarbonFileFilter) and reduce the number of namenode rpc

QiangCai:
Reduce the number of namenode rpc:
**1 Optimize HDFSCarbonFile.listFiles(CarbonFileFilter)**
  Firstly, call listFiles() to list all files;
  Secondly, use CarbonFileFilter on files
  Similar to FileSystem.listStatus implement of Hadoop in client side.
   filesystem
   **Disadvantage of old code:**
   a) use new HDFSCarbonFile(path) to create CarbonFile instance, this will send a namenode rpc for each path
   b) convert CarbonFileFilter to PathFilter
   **Advantage of new code:**
   a) change to use new HDFSCarbonFile(listStatus[i]) in listFiles() method to create CarbonFile instance, this will avoid rpc request.
   b) directly use CarbonFileFilter on files
   

**2 check dictionary file whether is exists or not**
    change to use listFiles to list all dictionary file, will reduce the number of rpc.

**Discussion:**
Because listFiles method will pull all file info to client side from namenode.  So I suggest that we should avoid to invoke listFiles method, if we just need to find few files. I think we can record file name in metadata file. What's your opinion?
I find list of calling listFiles in carbon, many util classes invoke this method, we should handle them  one by one.
listfiles


jarray888:
it will help to increase the cocurrent query. @kumarvishal09 ,please help to confirm .

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -297,9 +298,38 @@ object GlobalDictionaryUtil extends Logging {
...
-    primDimensions.zipWithIndex.foreach{f =>
-      dictFilePaths(f._2) = carbonTablePath.getDictionaryFilePath(f._1.getColumnId)
-      dictFileExists(f._2) = CarbonUtil.isFileExists(dictFilePaths(f._2))
+    val fileType = FileFactory.getFileType(dictfolderPath)
+    val metadataDirectory = FileFactory.getCarbonFile(dictfolderPath, fileType)
+    metadataDirectory.exists match {
+      case true =>
+        val carbonFiles = metadataDirectory.listFiles(new CarbonFileFilter {
+          @Override def accept(pathname: CarbonFile): Boolean = {
+            (!pathname.isDirectory) && pathname.getName.endsWith(CarbonTablePath.DICTIONARY_EXT)
```
gvramana: Dont expose, DICTIONARY as public, instead add isDictionaryFile function to CarbonTablePath

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -297,9 +298,38 @@ object GlobalDictionaryUtil extends Logging {
...
+          }
+        })
+        val fileNamesMap = new HashMap[String, Int]
+        for (i <- 0 until carbonFiles.length) {
+          fileNamesMap.put(carbonFiles(i).getName, i)
+        }
+        primDimensions.zipWithIndex.foreach { f =>
```
gvramana: please write comment about logic

kumarvishal09:
@QiangCai have to tested this code in hdfs?? how much benefit you are getting ??

Hexiaoqiao:
+1, it's a good optimization. i think the performance improvement is related with the number of files which is requests, and this belongs to HDFS, so shall we postpone the performance testing or is it necessary? pls correct me if i'm wrong. @kumarvishal09

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: fixed

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
QiangCai: done

QiangCai:
@gvramana please continue review

gvramana:
LGTM

PR #670, opened 2016-06-16
[Issue 618]Supported Spark 1.6 in Carbondata

ravipesala:
Spark 1.6 changes approach of aggregation and changed the interfaces. So all old implementations need to be either rewritten or deleted in this PR.

MeasureAggregator object is no longer supported in Spark 1.6 as UDT. As Spark 1.6 is completely based on UnSafeRow and codegen. So it try to read only the sqlType of UDT.
So approach is changed to avoid usage of MeasureAggregator objects in Spark side aggregation.

Distinct aggregation like count-distinct or sum-distinct is no longer supported to push down to spark layer. Spark changes the plan completely to support distinct count in its way, so carbon cannot get a chance to push down distinct aggregation.

Old PR link https://github.com/HuaweiBigData/carbondata/pull/624


ravipesala:
@jackylk I have created PR on spark16 branch

jackylk:
3 testcases are failing, please fix them.

ravipesala:
@jackylk those test cases are related to aggregation pushdown. So if we are going to remove aggregation push down then it will be of no use even though after fixing it.

ravipesala:
@jackylk Can I remove the push down code in this PR?

chenliang613:
@ravipesala   i suggest open a new PR to handle invalid aggregation code, and discuss it separately. please don't mix together.

jackylk:
@ravipesala thanks for working on this. I will do a code clean up in another PR.

PR #671, opened 2016-06-16
[Bug] Load status getting overridden in case one cluster fails to load the data

manishgupta88:
**Issue:** In case if data load fails on one cluster node and passes on other cluster node then as a load status a map with key as partition id and value as load status is end from the executor but currently partitioning is not supported. Due to this the load status when combined from all executor nodes is getting overridden and data load is passing sometimes even in case of failure on one node.

**Fix:** Make the key unique so that status does not get overriden


gvramana:
LGTM

PR #681, opened 2016-06-17
[Bug] Set ignore whitespace option to false for ignoring leading and trailing whitespace in data

manishgupta88:
To ignore trmimming of leading and trailing whitespace the option for ignoring leading and trailing whitespace need to be set to false while create dataframe using univocity library


manishgupta88:
@QiangCai ...kindly review and merge

QiangCai:
LGTM

PR #684, opened 2016-06-17
Refactored core package and fixed all testcases

ravipesala:
optimize all aggregate related code from core package and fixed all test cases.


ravipesala:
@jackylk Please review and merge

core/src/main/java/org/carbondata/query/carbon/aggregator/impl/ListBasedResultAggregator.java
```diff
@@ -114,54 +124,69 @@ public int aggregateData(AbstractScannedResult scannedResult) {
...
   }
 
   /**
    * Below method will used to get the result
-   *
-   * @param aggregated result
    */
-  @Override public Result<List<ListBasedResultWrapper>> getAggregatedResult() {
-    Result<List<ListBasedResultWrapper>> result = new ListBasedResult();
+  @Override public Result<List<ListBasedResultWrapper>, Object> getAggregatedResult() {
```
kumarvishal09: ListBasedResultWrapper is already holding measure and key array then why this object is required ??

core/src/main/java/org/carbondata/query/carbon/aggregator/impl/ListBasedResultAggregator.java
```diff
@@ -68,14 +65,6 @@
   private List<ListBasedResultWrapper> listBasedResult;
 
   /**
-   * aggreagtor used
-   *
-   * @TODO no need to create this
-   * @TODO need to handle in some other way
-   */
-  private MeasureAggregator[] blockAggregator;
```
kumarvishal09: in case of detail query we are using measure aggregator for just wrapping the measure value and data aggregator for fill the value ....now here we will hold in an object ...what benefit you are getting as both are object

core/src/main/java/org/carbondata/query/carbon/aggregator/impl/MapBasedResultAggregator.java
```diff
@@ -126,11 +126,12 @@ public MapBasedResultAggregator(BlockExecutionInfo tableBlockExecutionInfos,
...
    * Below method will used to get the result
-   *
-   * @param aggregated result
    */
-  @Override public Result<Map<ByteArrayWrapper, MeasureAggregator[]>> getAggregatedResult() {
-    Result<Map<ByteArrayWrapper, MeasureAggregator[]>> result = new MapBasedResult();
+  @Override
+  public Result<Map<ByteArrayWrapper, MeasureAggregator[]>,
+      MeasureAggregator[]> getAggregatedResult() {
+    Result<Map<ByteArrayWrapper, MeasureAggregator[]>, MeasureAggregator[]> result =
```
kumarvishal09: why Result<Map<ByteArrayWrapper, MeasureAggregator[]>, MeasureAggregator[]>  why two measure aggregator is required ??

core/src/main/java/org/carbondata/query/carbon/executor/QueryExecutorFactory.java
```diff
@@ -38,39 +34,8 @@
...
-    else if (queryModel.isDetailQuery() && queryModel.getSortDimension().size() > 0) {
-      LOGGER.info("Detail with order by query: ");
-      return new DetailWithOrderByQueryExecutor();
-    } else if (queryModel.isForcedDetailRawQuery()) {
-      return new DetailRawRecordQueryExcecutor();
+    if (queryModel.isForcedDetailRawQuery()) {
+      return new DetailRawRecordQueryExecutor();
     } else {
       // detail query
       LOGGER.info("Detail query: ");
```
jackylk: both are detail query, I think we can remove this comment and log

core/src/main/java/org/carbondata/query/carbon/aggregator/impl/ListBasedResultAggregator.java
ravipesala: It was added to satisfy current interfaces. It will be refactored later

core/src/main/java/org/carbondata/query/carbon/aggregator/impl/ListBasedResultAggregator.java
ravipesala: Measure aggregator object need to be created and also it has some state so better use Object

core/src/main/java/org/carbondata/query/carbon/aggregator/impl/MapBasedResultAggregator.java
ravipesala: It is also to satisfy current interfaces. Can be refactored later

core/src/main/java/org/carbondata/query/carbon/executor/QueryExecutorFactory.java
ravipesala: ok

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDropAggregateTableRDD.scala
```diff
@@ -18,23 +18,23 @@
...
 import org.apache.spark.sql.execution.command.Partitioner
 
-import org.carbondata.query.scanner.impl.{CarbonKey, CarbonValue}
-import org.carbondata.spark.KeyVal
+import org.carbondata.spark.Value
 import org.carbondata.spark.util.CarbonQueryUtil
 
 
-class CarbonDropAggregateTableRDD[K, V](
+class CarbonDropAggregateTableRDD[V: ClassTag](
```
jackylk: I think we can remove it as we do not support aggregate table currently

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDropCubeRDD.scala
```diff
@@ -17,21 +17,22 @@
...
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.execution.command.Partitioner
 
-import org.carbondata.query.scanner.impl.{CarbonKey, CarbonValue}
-import org.carbondata.spark.KeyVal
+import org.carbondata.spark.Value
 import org.carbondata.spark.util.CarbonQueryUtil
 
-class CarbonDropCubeRDD[K, V](
+class CarbonDropCubeRDD[V: ClassTag](
```
jackylk: please rename it to CarbonDropTableRDD

ravipesala:
@jackylk Handled review comments. Please review and merge

chenliang613:
LGTM.
Optimize aggregation module,make better integration with processing engine(Apache Spark).

jackylk:
LGTM

PR #690, opened 2016-06-18
Remove AbstractQueryExecutor

jackylk:
After aggregator refactory in #684, `CarbonAbstractQueryExecutor` is of no use, and thus be removed in this PR.


ravipesala:
LGTM

PR #691, opened 2016-06-18
Optimized detail query flow and cleanup

ravipesala:
Optimized detail query flow and cleanup is done here. It enables the data to return in proper batches.


ravipesala:
@jackylk . Please review and merge.

core/src/main/java/org/carbondata/query/carbon/aggregator/ScannedResultCollector.java
```diff
@@ -24,20 +24,20 @@
 /**
  * Interface which will be used to aggregate the scan result
  */
-public interface ScannedResultAggregator {
+public interface ScannedResultCollector {
```
jackylk: I think package name also can be changed

core/src/main/java/org/carbondata/query/carbon/result/iterator/DetailQueryResultIterator.java
```diff
@@ -42,48 +40,53 @@
 public class DetailQueryResultIterator extends AbstractDetailQueryResultIterator {
 
   /**
-   * LOGGER.
-   */
-  private static final LogService LOGGER =
-      LogServiceFactory.getLogService(DetailQueryResultIterator.class.getName());
```
jackylk: logger is not required?

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
```diff
@@ -60,101 +62,73 @@
...
+    this.fileReader = FileFactory.getFileHolder(
+        FileFactory.getFileType(queryModel.getAbsoluteTableIdentifier().getStorePath()));
     intialiseInfos();
   }
 
-  private void intialiseInfos() {
-    this.totalNumberBlockletPerSlice = new long[blockExecutionInfos.size()];
-    this.numberOfBlockletExecutedPerBlock = new long[blockExecutionInfos.size()];
-    int index = -1;
+  protected void intialiseInfos() {
```
jackylk: why is it protected? planning to add subclass to overwrite it?

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
```diff
@@ -60,101 +62,73 @@
...
-    return currentCounter < totalNumberOfNode;
+    if ((dataBlockProcessor != null && dataBlockProcessor.hasNext()) || nextBatch) {
+      return true;
+    } else {
+      dataBlockProcessor = getDataBlockProcessor();
+      while (dataBlockProcessor != null) {
+        if (dataBlockProcessor.hasNext()) {
+          return true;
+        }
+        dataBlockProcessor = getDataBlockProcessor();
```
jackylk: It is a Iterator, the function name and variable name is misleading

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
ravipesala: I tried to subclass but removed later. I will change it

core/src/main/java/org/carbondata/query/carbon/result/iterator/DetailQueryResultIterator.java
```diff
@@ -42,48 +40,53 @@
...
     }
+    return result;
+  }
+
+  private Future<BatchResult> execute() {
+    return execService.submit(new Callable<BatchResult>() {
+      @Override public BatchResult call() throws QueryExecutionException {
+        BatchResult batchResult;
+        Result next = dataBlockProcessor.next();
+        if (null != next) {
```
jackylk: this if check is redundant

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
ravipesala: I will update

core/src/main/java/org/carbondata/query/carbon/result/iterator/DetailQueryResultIterator.java
ravipesala: I will remove it

core/src/main/java/org/carbondata/query/carbon/result/preparator/impl/RawQueryResultPreparatorImpl.java
```diff
@@ -55,28 +60,64 @@ public RawQueryResultPreparatorImpl(QueryExecutorProperties executerProperties,
...
-      row[0] = scannedResult.getKey();
-      if(value != null) {
-        System.arraycopy(value, 0, row, 1, msrSize);
+    if (queryModel.isRawBytesDetailQuery()) {
+      while (scannedResult.hasNext()) {
+        value = scannedResult.getValue();
+        row = new Object[msrSize + 1];
+        row[0] = scannedResult.getKey();
+        if (value != null) {
+          System.arraycopy(value, 0, row, 1, msrSize);
```
jackylk: please add a assertion to compare value length and msrSize

ravipesala: But it is inside loop , and as well as we can guarantee that if `value` is not null then it would be  of size `msrSize`. Please check class `ListBasedResultCollector.collectData`, here we always create value of size equals to `msrSize`. So I guess this check would be redundant.

core/src/main/java/org/carbondata/query/carbon/result/preparator/impl/RawQueryResultPreparatorImpl.java
```diff
@@ -55,28 +60,64 @@ public RawQueryResultPreparatorImpl(QueryExecutorProperties executerProperties,
...
+              .getKeyArray(key.getDictionaryKey(), querySchemaInfo.getMaskedByteIndexes());
+          int noDictionaryColumnIndex = 0;
+          for (int i = 0; i < dimSize; i++) {
+            if (!queryDimensions[i].getDimension().hasEncoding(Encoding.DICTIONARY)) {
+              row[order[i]] = DataTypeUtil.getDataBasedOnDataType(
+                  new String(key.getNoDictionaryKeyByIndex(noDictionaryColumnIndex++)),
+                  queryDimensions[i].getDimension().getDataType());
+            } else if (queryDimensions[i].getDimension().hasEncoding(Encoding.DIRECT_DICTIONARY)) {
+              DirectDictionaryGenerator directDictionaryGenerator =
+                  DirectDictionaryKeyGeneratorFactory.getDirectDictionaryGenerator(
```
jackylk: it could return null if it is not timestamp type

ravipesala: Yes, as of now direct_dictionary is enabled only for timestamp

core/src/main/java/org/carbondata/query/carbon/result/preparator/impl/RawQueryResultPreparatorImpl.java
jackylk: Yes, so by design the size should be msrSize, but since it is returned by an interface call, I am suggesting to add an assertion to check it, but not if check.

core/src/main/java/org/carbondata/query/carbon/result/preparator/impl/RawQueryResultPreparatorImpl.java
jackylk: so there should be a null check before using `directionDictioanryGenerator`

ravipesala: ok

core/src/main/java/org/carbondata/query/carbon/result/preparator/impl/RawQueryResultPreparatorImpl.java
ravipesala: ok. Added

ravipesala:
@jackylk  Fixed review comments, please review and merge.

jackylk:
LGTM

PR #692, opened 2016-06-18
Refactor org.carbondata.query package

jackylk:
1. Remove unused classes
2. Move packages to make it more modular
3. Rename the package to org.carbondata.scan


ravipesala:
Packages looks more arranged now  :- )

PR #694, opened 2016-06-19
Add spark version print in example

jackylk:
It prints spark version in example while creating the context, as we will support multiple spark version.


ravipesala:
LGTM

PR #695, opened 2016-06-19
Fixed data loading failure in Spark1.6 spark shell

ravipesala:
It seems spark1.6 URL classloader server is buggy, it returns `META-INF/vfs-providers.xml` two times from jar file even though it is only one file. Because of this reason commons-vfs is failing. Any way we don't required this xml file to read as we are already handling in code.
So I am excluding it in assembly file. It works fine now both in Spark 1.5 and 1.6.


jackylk:
LGTM

PR #700, opened 2016-06-20
remove uneeded addProperty code

Zhangshunyu:
this property "carbon.table.split.partition.enable" is never used now.


chenliang613:
In the future, may support this feature, so keep the comment is ok.

PR #712, opened 2016-06-21
[CARBONDATA-96] Make zookeeper lock as default if zookeeper url is configured.

ravikiran23:
make the lock type as zookeeper if zookeeper URL is present in the spark conf.
if spark.deploy.zookeeper.url property is set in spark-default.conf then need to take the zookeeper locking.


ravikiran23:
retest this please

gvramana:
retest this please

PR #713, opened 2016-06-21
[Issue 728]Column group with Dicitonary Exclude

ashokblend:
1.Corrected Encoding of chunk while dataloading 
2. Fixed min max issue occured due to one to one one mapping distrubance.
3. Fixed Query execution issue

### DIFF
--- a/core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
+++ b/core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
@@ -21,10 +21,12 @@ package org.carbondata.core.carbon.datastore.block;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 import org.carbondata.core.carbon.metadata.datatype.DataType;
 import org.carbondata.core.carbon.metadata.encoder.Encoding;
@@ -37,6 +39,7 @@ import org.carbondata.core.keygenerator.columnar.ColumnarSplitter;
 import org.carbondata.core.keygenerator.columnar.impl.MultiDimKeyVarLengthVariableSplitGenerator;
 import org.carbondata.core.keygenerator.mdkey.MultiDimKeyVarLengthGenerator;
 import org.carbondata.core.util.CarbonUtil;
+import org.carbondata.core.vo.ColumnGroupModel;
 
 import org.apache.commons.lang3.ArrayUtils;
 
@@ -85,6 +88,12 @@ public class SegmentProperties {
   private Map<Integer, Integer> dimensionOrdinalToBlockMapping;
 
   /**
+   * a block can have multiple columns. This will have block index as key
+   * and all dimension participated in that block as values
+   */
+  private Map<Integer, Set<Integer>> blockTodimensionOrdinalMapping;
+
+  /**
    * mapping of measure column to block to in file this will be used while
    * reading the block in a file
    */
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
```diff
@@ -584,11 +638,63 @@ public int getNumberOfNoDictionaryDimension() {
...
+   * @return colGroupModel
+   */
+  public ColumnGroupModel getColumnGroupModel() {
+    return this.colGroupModel;
+  }
+
+  public boolean containsEncoding(int blockIndex, Encoding encoding) {
+    Set<Integer> dimOrdinals = blockTodimensionOrdinalMapping.get(blockIndex);
+    //column groups will always have dictionary encoding
+    if (dimOrdinals.size() > 1 && Encoding.DICTIONARY.equals(encoding)) {
```
kumarvishal09: no need to check equals reference check will also work fine as its a enum

core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
```diff
@@ -74,6 +77,17 @@
   private int[] dimColumnsCardinality;
 
   /**
+   * Cardinality of dimensions, including no dictionary column
+   * no dictionary column cardinality will be -1
+   */
+  private int[] dictIncludeColCardinality;
```
kumarvishal09: there are lots of redundant information we are storing in this class , for example  dimColumnsCardinality and dictIncludeColCardinality only difference i can find is dimColumnsCardinality  is having -1 for no dictionary column, is it really required to store dictIncludeColCardinality ?

core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
```diff
@@ -74,6 +77,17 @@
...
+
+  /**
+   * Column schema
+   */
+  private List<ColumnSchema> columnSchemas;
```
kumarvishal09: column schema is already present in the CarbonTable why we are storing it again segment properties ??

core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
```diff
@@ -114,6 +134,12 @@
   private Map<Integer, KeyGenerator> columnGroupAndItsKeygenartor;
 
   /**
+   * column group key generator dimension index will not be same as dimension ordinal
+   * This will have mapping with ordinal and keygenerator or mdkey index
+   */
+  private Map<Integer, Map<Integer, Integer>> columnGroupOrdinalToMdkeymapping;
```
kumarvishal09: is this column group to its dimension to mdkey mapping (complete mdkey mapping)? or column group mdkey mapping ? if complete mdkey mapping then this we can  calculate in run time why we need to persist this information?

core/src/main/java/org/carbondata/core/util/CarbonUtil.java
```diff
@@ -1682,5 +1682,43 @@ public static void clearDictionaryCache(Dictionary dictionary) {
     }
   }
 
+  public static int[] getFormattedCardinality(int[] dictionaryColumnCardinality,
```
kumarvishal09: please add some comments

core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
ashokblend: since we have keygenerator for each column groups. This mapping is for dimension ordinal to mdkey mapping of given column group key generator

core/src/main/java/org/carbondata/core/carbon/datastore/block/SegmentProperties.java
ashokblend: As we know this list of columnSchema have list of all dimension including chldrens of any dimension in case of complex type. If we want to use CarbonTable , then we need to evaluate this list every time.

ashokblend:
@kumarvishal09 From SegmentProperties i have removed below varibles
1.storing flattened columnschem
2.storing dictionaryIncludeCardinality.
Please check.

kumarvishal09:
retest this please

kumarvishal09:
LGTM

PR #718, opened 2016-06-22
[issue-717] select id from table_x where salary is not null; returnin…

mohammadshahidkhan:
…g the empty result even though null data is available

Problem  and solution:
For storing the null value we are giving the value  (minValue-1) and  mainting the null indexes in bitset,
While applying the filter we were not  comparison with actual null value instead doing with the minValue-1.
Now while creating the row instances, assigning the null value at the null index.


ravipesala:
LGTM

PR #719, opened 2016-06-22
[BUG] When data load is partially successful, log should be written as data load partially successful

nareshpr:


gvramana:
LGTM

PR #720, opened 2016-06-22
Refactor carbon-core module for code clean up

jackylk:
code clean up as following:
1. remove unused class and function in carbon-core module
2. move LCM related packages from carbon-core to carbon-processing module


ravipesala:
@jackylk Following also can be refactored
1. There are two `metadata` folders are present in core module and some unused class is also present in that folder. Better merge to one folder.
2. There are 2 `DataTypeUtil` classes are there in core package. Can be merged as one.
3. Package `integration/spark/org/carbondata/spark/query/metadata` can be deleted.
4. There are 2 classes 'DataType' . May be one can renamed to avoid confusion.
5. `DataSetUtil` can be removed.
6. In package `org.carbondata.core.carbon`. Again `carbon` is required in package name? May be we can remove `carbon` in package name

jackylk:
Yes, I can create other PR to modify them.

ravipesala:
Ok. LGTM.

PR #683, opened 2016-06-17
[Bug] Row level scanning filters while filtering invalid member giving exception

sujith71955:
Row level scanning filters while filtering invalid member giving exception, as per the hive compatibility
empty rows will be displayed .

Problem:
While performng Row level scanning to apply Multi dimension filters, if any non parseable data present
filter layer was throwing out run time exception.

Solution:
Filter layer shall provide empty resultset in this scenario, if any NumberFormatException happens while comparison, catch the same and pass empty resultset.

PR #721, opened 2016-06-22
[CARBONDATA-13] Direct surrogate key range filters not able to prune blocks

sujith71955:
Direct surrogate key range filters not able to prune blocks, Now the support is added for direct surrogates column also. previously row level filter evaluation was happening for direct dictionary columns in case of range filters, this will reduce the performance also for such queries which is having timestamp data with range filters applied

Note: The above PR has to be rebased with PR #683 , #683 PR needs to be merged first.
.

### DIFF
--- a/core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
+++ b/core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
@@ -129,7 +129,19 @@ public class RowLevelRangeGrtThanFiterExecuterImpl extends RowLevelFilterExecute
       // from the next element which is greater than filter member.
       if (start < 0) {
         start = -(start + 1);
+        if (start == numerOfRows) {
+          start = start - 1;
+        }
+        // Method will compare the tentative index value after binary search, this tentative
+        // index needs to be compared by the filter member if its > filter then from that
+        // index the bitset will be considered for filtering process.
+        if (ByteUtil
+            .compare(filterValues[i], dimensionColumnDataChunk.getChunkData(columnIndex[start]))
+            > 0) {
+          start = start + 1;
+        }
       }
+
       last = start;
       for (int j = start; j < numerOfRows; j++) {
         bitSet.set(columnIndex[j]);
@@ -170,9 +182,18 @@ public class RowLevelRangeGrtThanFiterExecuterImpl extends RowLevelFilterExecute
             (FixedLengthDimensionDataChunk) dimensionColumnDataChunk, filterValues[k], numerOfRows);
         if (start < 0) {
           start = -(start + 1);
+          if (start == numerOfRows) {
+            start = start - 1;
+          }
+          // Method will compare the tentative index value after binary search, this tentative
+          // index needs to be compared by the filter member if its > filter then from that
+          // index the bitset will be considered for filtering process.
+          if (ByteUtil.compare(filterValues[k], dimensionColumnDataChunk.getChunkData(start)) > 0) {
+            start = start + 1;
+          }
         }
         last = start;
-        for (int j = start; j <= numerOfRows; j++) {
+        for (int j = start; j < numerOfRows; j++) {
           bitSet.set(j);
           last++;
         }
--- a/core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
+++ b/core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
@@ -22,20 +22,30 @@ import java.util.BitSet;
 import java.util.List;
 
 import org.carbondata.core.carbon.AbsoluteTableIdentifier;
+import org.carbondata.core.carbon.datastore.block.SegmentProperties;
+import org.carbondata.core.carbon.datastore.chunk.DimensionColumnDataChunk;
+import org.carbondata.core.carbon.datastore.chunk.impl.FixedLengthDimensionDataChunk;
+import org.carbondata.core.carbon.metadata.encoder.Encoding;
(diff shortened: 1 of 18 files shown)


sujith71955:
Please review

sujith71955:
retest this please

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
```diff
@@ -63,4 +72,138 @@ public RowLevelRangeGrtThanFiterExecuterImpl(
...
+        start = -(start + 1);
+        if (start == numerOfRows) {
+          start = start - 1;
+        }
+        // Method will compare the tentative index value after binary search, this tentative
+        // index needs to be compared by the filter member if its > filter then from that
+        // index the bitset will be considered for filtering process.
+        if (ByteUtil
+            .compare(filterValues[i], dimensionColumnDataChunk.getChunkData(columnIndex[start]))
+            > 0) {
```
gvramana: Other than greater than logic complete logic is same across filter. Separate comparison logic and move remaining to base class

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
```diff
@@ -63,4 +72,138 @@ public RowLevelRangeGrtThanFiterExecuterImpl(
...
+            (FixedLengthDimensionDataChunk) dimensionColumnDataChunk, filterValues[k], numerOfRows);
+        if (start < 0) {
+          start = -(start + 1);
+          if (start == numerOfRows) {
+            start = start - 1;
+          }
+          // Method will compare the tentative index value after binary search, this tentative
+          // index needs to be compared by the filter member if its > filter then from that
+          // index the bitset will be considered for filtering process.
+          if (ByteUtil.compare(filterValues[k], dimensionColumnDataChunk.getChunkData(start)) > 0) {
```
gvramana: Other than greater than logic complete logic is same across filter. Separate comparison logic and move remaining to base class

core/src/main/java/org/carbondata/query/filter/resolver/RowLevelRangeFilterResolverImpl.java
```diff
@@ -191,6 +206,29 @@ public void resolve(AbsoluteTableIdentifier absoluteTableIdentifier) {
     }
   }
 
+  private List<Integer> getDirectSurrogateValues(ColumnExpression columnExpression) {
+    List<ExpressionResult> listOfExpressionResults = new ArrayList<ExpressionResult>(20);
+    DirectDictionaryGenerator directDictionaryGenerator = DirectDictionaryKeyGeneratorFactory
+        .getDirectDictionaryGenerator(columnExpression.getDimension().getDataType());
+
+    if (this.getFilterExpression() instanceof BinaryConditionalExpression) {
```
gvramana: Why check is only for BinaryConditionExpression, how about others?
also write comment

integration/spark/src/test/scala/org/carbondata/spark/testsuite/filterexpr/FilterProcessorTestCase.scala
```diff
@@ -85,6 +85,13 @@ class FilterProcessorTestCase extends QueryTest with BeforeAndAfterAll {
       Seq(Row(4), Row(6))
     )
   }
+  
+    test("Between  filter") {
+    checkAnswer(
+      sql("select date from filtertestTablesWithNull " + " where date between '2014-01-20 00:00:00' and '2014-01-28 00:00:00'"),
+      Seq(Row(Timestamp.valueOf("2014-01-21 00:00:00")), Row(Timestamp.valueOf("2014-01-22 00:00:00")))
+    )
+  }
```
gvramana: Add testcases for >,>=,<,<= cases for Direct dictionary and no-Dictionary cases also

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
sujith71955: Logic is written a loop where we are iterating filters and trying to find range, in this case also certain decisions needs to be taken in based on inverted index(present or not), so moving to base class means again comparison logic has to be added with inverted and non inverted which will make code bit complex to read/understand, currently the comparison logic is happening based on each type of Range filters. What you suggest.

integration/spark/src/test/scala/org/carbondata/spark/testsuite/filterexpr/FilterProcessorTestCase.scala
sujith71955: ok

core/src/main/java/org/carbondata/query/filter/resolver/RowLevelRangeFilterResolverImpl.java
sujith71955: scenario comes under range filter which is nothing but >,<,>=,<= which are nothing but binary filter condition

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
sujith71955: will be fix part of #766

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeGrtThanFiterExecuterImpl.java
sujith71955: will be fix part of #766

PR #722, opened 2016-06-22
Supported carbon spark sql cli and carbon spark shell in carbondata to simplify operations for first time users

ravipesala:
This PR makes the things more simplified for first time users. And our Quick start guide will be pretty neat.
**Carbon Spark SQL CLI**
 New sh file is added in carbon `bin/carbon-spark-sql`. User just needs to execute this file and execute the queries directly in the CLI. It works as like spark-sql cli.
No need set any classpath here.

Example : 

`carbondata/bin>./carbon-spark-sql`

`spark-sql> select * from table1`

**Carbon Spark shell**
New sh file is added in carbon `bin/carbon-spark-shell`. User can execute file to open the spark shell. User no need set any class path here.

Example 

`carbondata/bin>./carbon-spark-shell`

`scala>import org.apache.spark.sql.CarbonContext`

`scala>val cc = new CarbonContext(sc, "./carbondata/store")`

### DIFF
--- a/bin/carbon-spark-shell
+++ b/bin/carbon-spark-shell
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Enter posix mode for bash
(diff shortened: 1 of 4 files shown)


chenliang613:
For Carbon Spark SQL CLI, suggest using carbon-spark-sql as prefix
example:
carbondata/bin>./carbon-spark-sql
carbon-spark-sql> select * from table1

chenliang613:
In my machine, both can't work fine, error msg as below:
1.carbon-spark-sql:
WARN  24-06 14:35:22,661 - Retrying creating default database after error: Error creating transactional connection factory
javax.jdo.JDOFatalInternalException: Error creating transactional connection factory
Caused by: org.datanucleus.exceptions.NucleusException: Attempt to invoke the "BONECP" plugin to create a ConnectionPool gave an error : The specified datastore driver ("com.mysql.jdbc.Driver") was not found in the CLASSPATH. Please check your CLASSPATH specification, and the name of the driver.

2.carbon-spark-shell:
WARN  24-06 14:37:33,343 - Retrying creating default database after error: Error creating transactional connection factory
javax.jdo.JDOFatalInternalException: Error creating transactional connection factory
WARN  24-06 14:37:33,400 - Failed to access metastore. This class should not accessed in runtime.
org.apache.hadoop.hive.ql.metadata.HiveException: java.lang.RuntimeException: Unable to instantiate org.apache.hadoop.hive.ql.metadata.SessionHiveMetaStoreClient

ravipesala:
Closing it.

PR #729, opened 2016-06-23
Fixed driver block loading performance issue

kumarvishal09:
1.Writing the block metadata in different file
2.Need to do the data load again if data is already written as new file is added 
3. It will improve the block loading in when block is getting loaded in driver as now we are opening less and reading very very less data . this will be usefull in case of filter query and count(*) query


Vimal-Das:
LGTM except for missing tests, @kumarvishal09 Please add some tests.

sujith71955:
retest this please

kumarvishal09:
@Vimal-Das this is existing flow and test cases are already present for this

PR #731, opened 2016-06-23
[Issue 730]Modified Describe formatted command

ashokblend:
Timestamp dimension is showing as Dictionary column. Correct to make it as Key Column


Vimal-Das:
LGTM

Vimal-Das:
Merged to master.

PR #733, opened 2016-06-23
Merge two DataTypeUtil classes

jackylk:
There are two DataTypeUtil in carbon-core module, merging them into 1 class.

### DIFF
--- a/core/src/main/java/org/carbondata/scan/collector/impl/ListBasedResultCollector.java
+++ b/core/src/main/java/org/carbondata/scan/collector/impl/ListBasedResultCollector.java
@@ -26,6 +26,7 @@ import org.carbondata.common.logging.LogServiceFactory;
 import org.carbondata.core.carbon.datastore.chunk.MeasureColumnDataChunk;
 import org.carbondata.core.carbon.metadata.datatype.DataType;
 import org.carbondata.core.keygenerator.KeyGenException;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/carbondata/scan/collector/impl/ListBasedResultCollector.java
```diff
@@ -161,7 +160,7 @@ private Object getMeasureData(MeasureColumnDataChunk dataChunk, int index, DataT
         default:
           msrVal = dataChunk.getMeasureDataHolder().getReadableDoubleValueByIndex(index);
       }
-      return DataTypeUtil.getMeasureDataBasedOnDataType(msrVal, dataType);
+      return org.carbondata.core.util.DataTypeUtil.getMeasureDataBasedOnDataType(msrVal, dataType);
```
ravipesala: Why not import `DataTypeUtil`? Please import in all other places also.

jackylk: will fix it. It is generated by IDE refactory function

ravipesala:
@jackylk can you raise these refactoring PR's on new apache git? we already migrated to apache git so all new PR can be raised to new git.

jackylk:
@ravipesala I think the new repo is not ready, we still need to use this repo for some days

PR #736, opened 2016-06-24
[BUG] Sort is not correct for Table blocks.

ravikiran23:
sorting logic of the table blocks was considering task no as string which is wrong. Need to convert it to int.


Vimal-Das:
LGTM, Merged to master.

PR #738, opened 2016-06-24
[Issue- CARBONDATA-7] fortify issue fixes, Fixed Explicit null derefe…

mohammadshahidkhan:
…renced ,Dereference null return value


mohammadshahidkhan:
ColumnGroupDataTypesTestCase	0.59 sec	0		0		7	-4	7	-4

negative test case is not failure its a diff with the previous build

ravipesala:
LGTM

PR #739, opened 2016-06-24
Removed unused properties from carbon

nareshpr:


nareshpr:
retest this please

core/src/main/java/org/carbondata/core/constants/CarbonCommonConstants.java
```diff
@@ -126,25 +78,9 @@
    */
   public static final String STORE_LOCATION_DEFAULT_VAL = "../unibi-solutions/system/carbon/store";
```
ravipesala: Change the path name `unibi-solutions`

conf/carbon.properties.template
```diff
@@ -0,0 +1,94 @@
+#################### Performance Configuration ##################
```
ravipesala: Pls remove molap.properties from `dev` folder.

nareshpr:
Fixed review comments

processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -88,12 +88,6 @@
   private static final LogService LOGGER =
       LogServiceFactory.getLogService(CarbonFactDataHandlerColumnar.class.getName());
   /**
-   * decimalPointers
```
ravipesala: Why this property is removed? why are you always taking default value? Why not let user configure this value? Any reason?

ravipesala:
LGTM

PR #740, opened 2016-06-24
[Issue - CARBONDATA-8] Using create table instead of create cube in all existing test cases

ManoharVanam:
1. Use create table instead of cube in all test cases
2. Remove unnecessary & duplicate  test cases


ravipesala:
LGTM

PR #741, opened 2016-06-24
[BUG] Failed to update table status on concurrent compaction and data load.

ravikiran23:
If the table status file needs to be updated then the table status lock should be acquired and done.
if the locking is failed because of other process locking then need to fail the complete Process.
If table status lock is failed to acquire then complete data load or the compaction should fail.
Correcting the local lock.
Adding the FT for compaction locking.

### DIFF
--- a/integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
+++ b/integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
@@ -48,6 +48,9 @@ import org.carbondata.core.datastorage.store.filesystem.CarbonFileFilter;
 import org.carbondata.core.datastorage.store.impl.FileFactory;
 import org.carbondata.core.datastorage.store.impl.FileFactory.FileType;
 import org.carbondata.core.load.LoadMetadataDetails;
+import org.carbondata.core.locks.CarbonLockFactory;
+import org.carbondata.core.locks.ICarbonLock;
+import org.carbondata.core.locks.LockUsage;
 import org.carbondata.core.util.CarbonProperties;
 import org.carbondata.core.util.CarbonUtil;
 import org.carbondata.core.util.CarbonUtilException;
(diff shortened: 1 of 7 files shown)


integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -687,51 +690,78 @@ public static String getLoadFolderPath(CarbonLoadModel loadModel, String cubeNam
    * This API will write the load level metadata for the loadmanagement module inorder to
    * manage the load and query execution management smoothly.
    */
```
gvramana: add comment for return type

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -687,51 +690,78 @@ public static String getLoadFolderPath(CarbonLoadModel loadModel, String cubeNam
...
-    loadMetadataDetails.setLoadName(String.valueOf(loadCount));
-    loadMetadataDetails.setLoadStartTime(startLoadTime);
-    LoadMetadataDetails[] listOfLoadFolderDetailsArray = null;
+    boolean status = false;
+
+    String metaDataFilepath =
+        loadModel.getCarbonDataLoadSchema().getCarbonTable().getMetaDataFilepath();
+
+    ICarbonLock carbonLock =
+        CarbonLockFactory.getCarbonLockObj(metaDataFilepath, LockUsage.TABLE_STATUS_LOCK);
```
gvramana: Add function getLock in status manager, which will create lockobject and give to user of status manager

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -791,12 +791,18 @@ object CarbonDataRDDFactory extends Logging {
...
+          val status = CarbonLoaderUtil
             .recordLoadMetadata(currentLoadCount,
               metadataDetails,
               carbonLoadModel,
               loadStatus,
-              loadStartTime)
+              loadStartTime
+            )
+          if (!status) {
+            val message = "Dataload failure"
```
gvramana: Give clear message why load has failed.

ravikiran23: fixed

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
ravikiran23: fixed

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
ravikiran23: fixed

gvramana:
LGTM

PR #742, opened 2016-06-24
[Issue - CARBONDATA-9] Fixed No lease issue on the bad record log file.

mohammadshahidkhan:
The issue was happening due to concurrent access of same file by three executors, if one of executor hold the lease on the file, the other two were not getting the lease, and ending up by throwing 
FileNotFoundException: No lease on filePath. 

Solution:
Adding the taskNo in the path to avoid concurrent access of same file.


ravipesala:
LGTM

PR #750, opened 2016-06-25
Dictionary creation is successful when distinct count of this column above high.cardinality.threshold 

foryou2030:
when we config high.cardinality.threshold=10000 in carbon.properties, it doesn't work.
in this pr, i have fixed this issue


QiangCai:
LGTM

PR #751, opened 2016-06-25
[Bug]skip escape char in kettle and fix quote in middle of value

eason-lyx:
in currectly flow,escape character will no skip in kettle,while dictionary generation will skip the escape character .and if the value no start with quote,it will no skip also.
 like "ab\c"->"abc",a"b\c"->a"b\c"

also,this pr fix the value like ab"c"d,which contain double quote in middle of value.

as suggesoin,we need to make the same parser in phase 1:global dictionary and phase 2:data loading.
using same parser:univocity parser. then all defects about different parser will no comes any more.


integration/spark/src/test/scala/org/carbondata/spark/testsuite/dataload/TestLoadDataWithHiveSyntax.scala
```diff
@@ -222,7 +222,24 @@ class TestLoadDataWithHiveSyntax extends QueryTest with BeforeAndAfterAll {
       .addProperty(CarbonCommonConstants.CARBON_TIMESTAMP_FORMAT, "dd-MM-yyyy")
     sql("DROP TABLE IF EXISTS t3")
   }
-  
+
+  test("test carbon table data loading with special character") {
```
Vimal-Das: Can you add some tests by passing different escape character in options and check whether the behavior is correct

Vimal-Das:
LGTM, Merged to Master.

PR #752, opened 2016-06-25
Added Validation for delete segments query, load start time parameter

ManoharVanam:
Added Validation for delete segments query, load start time parameter

### DIFF
--- a/core/src/main/java/org/carbondata/core/load/LoadMetadataDetails.java
+++ b/core/src/main/java/org/carbondata/core/load/LoadMetadataDetails.java
@@ -20,6 +20,13 @@
 package org.carbondata.core.load;
 
 import java.io.Serializable;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import org.carbondata.common.logging.LogService;
+import org.carbondata.common.logging.LogServiceFactory;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/carbondata/core/load/LoadMetadataDetails.java
```diff
@@ -126,6 +139,33 @@ public String getLoadStartTime() {
...
+  /**
+   * returns load start time as long value
+   * @param loadStartTime
+   * @return
+   */
+  private Long getTimeStamp(String loadStartTime) {
+    if (loadStartTime.isEmpty()) {
+      return null;
+    }
+    SimpleDateFormat parser = new SimpleDateFormat(CarbonCommonConstants.CARBON_TIMESTAMP);
```
ravipesala: it could be static and class level attribute

ravipesala:
LGTM except one small review comment. Please fix it I will merge it.

gvramana:
LGTM

core/src/main/java/org/carbondata/core/load/LoadMetadataDetails.java
ManoharVanam: fixed.

PR #754, opened 2016-06-26
[Bug] Timestamp with different format compare to spark date format is failing while query

sujith71955:
Timestamp with different format compare to spark date format will fail to provide filter result. this is because while querying the engine is trying to get the direct surrogate based in user defined time format in properties file.

### DIFF
--- a/core/src/main/java/org/carbondata/core/keygenerator/directdictionary/timestamp/TimeStampDirectDictionaryGenerator.java
+++ b/core/src/main/java/org/carbondata/core/keygenerator/directdictionary/timestamp/TimeStampDirectDictionaryGenerator.java
@@ -40,12 +40,6 @@ import static org.carbondata.core.keygenerator.directdictionary.timestamp.TimeSt
 public class TimeStampDirectDictionaryGenerator implements DirectDictionaryGenerator {
 
   /**
-   * Logger instance
-   */
-  private static final LogService LOGGER =
-      LogServiceFactory.getLogService(TimeStampDirectDictionaryGenerator.class.getName());
-
-  /**
    * The value of 1 unit of the SECOND, MINUTE, HOUR, or DAY in millis.
    */
   public static final long granularityFactor;
@@ -55,6 +49,11 @@ public class TimeStampDirectDictionaryGenerator implements DirectDictionaryGener
    * customized the start of position. for example "January 1, 2000"
    */
   public static final long cutOffTimeStamp;
+  /**
+   * Logger instance
+   */
+  private static final LogService LOGGER =
+      LogServiceFactory.getLogService(TimeStampDirectDictionaryGenerator.class.getName());
 
   /**
    * initialization block for granularityFactor and cutOffTimeStamp
@@ -119,6 +118,27 @@ public class TimeStampDirectDictionaryGenerator implements DirectDictionaryGener
         .equals(CarbonCommonConstants.MEMBER_DEFAULT_VAL)) {
       return 1;
     }
+    return getDirectSurrogateForMember(memberStr, timeParser);
+  }
+
+  /**
+   * The method take member String as input and converts
+   * and returns the dictionary key
+   *
+   * @param memberStr date format string
+   * @return dictionary value
+   */
+  public int generateDirectSurrogateKey(String memberStr, String format) {
+    SimpleDateFormat timeParser = new SimpleDateFormat(format);
+    timeParser.setLenient(false);
+    if (null == memberStr || memberStr.trim().isEmpty() || memberStr
+        .equals(CarbonCommonConstants.MEMBER_DEFAULT_VAL)) {
+      return 1;
+    }
+    return getDirectSurrogateForMember(memberStr, timeParser);
+  }
+
+  private int getDirectSurrogateForMember(String memberStr, SimpleDateFormat timeParser) {
     Date dateToStr = null;
     try {
       dateToStr = timeParser.parse(memberStr);
--- a/core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
+++ b/core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
@@ -243,9 +243,12 @@ public class ExpressionResult implements Comparable<ExpressionResult> {
     try {
       switch (this.getDataType()) {
         case StringType:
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/carbondata/core/keygenerator/directdictionary/timestamp/TimeStampDirectDictionaryGenerator.java
```diff
@@ -93,9 +92,6 @@
         Date dateToStr = timeParser.parse(cutOffTimeStampString);
         cutOffTimeStampLocal = dateToStr.getTime();
       } catch (ParseException e) {
-        LOGGER.warn("Cannot convert" + cutOffTimeStampString
```
ravipesala: why this log is removed?

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
```diff
@@ -243,9 +243,8 @@ public Long getTime() throws FilterIllegalMemberException {
     try {
       switch (this.getDataType()) {
         case StringType:
-          SimpleDateFormat parser = new SimpleDateFormat(CarbonProperties.getInstance()
-              .getProperty(CarbonCommonConstants.CARBON_TIMESTAMP_FORMAT,
-                  CarbonCommonConstants.CARBON_TIMESTAMP_DEFAULT_FORMAT));
+          SimpleDateFormat parser = new SimpleDateFormat(CarbonCommonConstants
+              .CARBON_TIMESTAMP_DEFAULT_FORMAT);
```
ravipesala: you mean spark always provides the dateformat in default format?

core/src/main/java/org/carbondata/core/keygenerator/directdictionary/timestamp/TimeStampDirectDictionaryGenerator.java
```diff
@@ -123,20 +120,27 @@
...
+        LOGGER.debug(
+            "Cannot convert " + memberStr + " to Time/Long type value. Value considered as null."
+                + e.getMessage());
+        dateToStr = null;
+      }
     }
     //adding +2 to reserve the first cuttOffDiff value for null or empty date
     if (null == dateToStr) {
-      return 1;
+      return -1;
```
ravipesala: It is supposed to be 1, why changed to -1. records will go to badrecords. Please change back

ravipesala:
Add both positive and negative testcase for this scenario.

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
sujith71955: query engine layer only supports yyyy-MM-dd HH:mm:ss date format no matter in which format the data is been stored, so while retrieving the direct surrogate value for
filter member first it should be converted in date form as per sparks supported date format and needs to retrieve time stamp for applying filters.User can load the data based on the format specified in carbon properties if provided

sujith71955:
All comments handled, Please merge the same. Thanks.

core/src/main/java/org/carbondata/core/keygenerator/directdictionary/timestamp/TimeStampDirectDictionaryGenerator.java
```diff
@@ -109,6 +108,7 @@
    *
    * @param memberStr date format string
    * @return dictionary value
+   * @throws FilterUnsupportedException
```
ravipesala: you are not throwing `FilterUnsupportedException` any where? why it is added to comment

core/src/main/java/org/carbondata/core/keygenerator/directdictionary/timestamp/TimeStampDirectDictionaryGenerator.java
```diff
@@ -119,6 +119,28 @@
...
+    return getDirectSurrogateForMember(memberStr, timeParser);
+  }
+
+  /**
+   * The method take member String as input and converts
+   * and returns the dictionary key
+   *
+   * @param memberStr date format string
+   * @return dictionary value
+   * @throws FilterUnsupportedException
```
ravipesala: same as above comment

core/src/main/java/org/carbondata/query/filter/resolver/resolverinfo/visitor/CustomTypeDictionaryVisitor.java
```diff
@@ -65,13 +66,15 @@ public void populateFilterResolvedInfo(DimColumnResolvedFilterInfo visitableObj,
 
   private DimColumnFilterInfo getDirectDictionaryValKeyMemberForFilter(
       AbsoluteTableIdentifier tableIdentifier, ColumnExpression columnExpression,
-      List<String> evaluateResultListFinal, boolean isIncludeFilter) {
+      List<String> evaluateResultListFinal, boolean isIncludeFilter)
+      throws FilterUnsupportedException {
```
ravipesala: why `FilterUnsupportedException` is required to throw here?

integration/spark/src/test/scala/org/carbondata/spark/testsuite/filterexpr/FilterProcessorTestCase.scala
```diff
@@ -40,10 +40,34 @@ class FilterProcessorTestCase extends QueryTest with BeforeAndAfterAll {
...
     sql("drop table if exists filtertestTablesWithNull")
+    sql("drop table if exists filterWithTimeStamp")
     sql("CREATE TABLE filtertestTables (ID int, date Timestamp, country String, " +
       "name String, phonetype String, serialname String, salary int) " +
         "STORED BY 'org.apache.carbondata.format'"
     )
+    
+     CarbonProperties.getInstance()
+        .addProperty(CarbonCommonConstants.CARBON_TIMESTAMP_FORMAT, "MM-dd-yyyy HH:mm:ss")
+     sql("CREATE CUBE filterWithTimeStamp DIMENSIONS (ID Integer,date Timestamp, country String, " +
```
ravipesala: don't use `CREATE CUBE`, use 'CREATE TABLE' command

sujith71955:
Done.

ravipesala:
LGTM

PR #755, opened 2016-06-27
[Bug] Exception thrown in sort and fact data writer step if all the records go to bad records

manishgupta88:
**Issue:**
1. In case all the records go to bad records in in surrogate key gen step, then also level cardinality file is getting written. Fact file writer gets initialized when level cardinality file is written and after that the process fails as there is no actual data leading to data load failure.
2. The sort key step throws null pointer exception as empty object row which has null value is put from surrogate step.

**Fix:**
1. Write level cardinality file only if there is at least one valid row in data.
2. If there is no valid data, then do call putRow method.


QiangCai:
LGTM

PR #756, opened 2016-06-27
Make inverted index can be configurable

Zhangshunyu:
Using DDL to determine which column not do inverted index, if the user doesn't specify for one column, the default value is true which means do inverted index.
For example:
```
(id Int, name String, city String)
TBLPROPERTIES('NO_INVERTED_INDEX'='name,city')
```


core/src/main/java/org/carbondata/core/datastorage/store/columnar/BlockIndexerStorageForNoInvertedIndex.java
```diff
@@ -0,0 +1,133 @@
...
+          start+=counter;
+          counter=1;
+          continue;
+        }
+        counter++;
+      }
+      map.add(start);
+      map.add(counter);
+      this.keyBlock = convertToKeyArray(list);
+      if(keyBlockInput.length == this.keyBlock.length) {
```
ManoharVanam: Pls Store keyBlockInput.length in one variable instead of getting everytime and use the same in in for loop

PR #757, opened 2016-06-27
fixed bug when high.cardinality.threshold == invalid number

foryou2030:
set high.cardinality.threshold as min value  when high.cardinality.threshold == invalid number


core/src/main/java/org/carbondata/core/util/CarbonProperties.java
```diff
@@ -256,10 +256,10 @@ private void validateHighCardinalityThreshold() {
       }
     } catch (NumberFormatException e) {
       LOGGER.info("The high cardinality threshold value \"" + highcardThresholdStr
```
gvramana: It should be logger.error

PR #760, opened 2016-06-27
[issue-CARBONDATA-15] filter query throwing error if the query applied over a table having no data.

mohammadshahidkhan:
Fixed https://issues.apache.org/jira/browse/CARBONDATA-15

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
@@ -80,11 +80,14 @@ class CarbonQueryRDD[K, V](
     val LOGGER = LogServiceFactory.getLogService(this.getClass.getName)
     // set filter resolver tree
     try {
-      var filterResolver = carbonInputFormat
-        .getResolvedFilter(job.getConfiguration, filterExpression)
-
-      CarbonInputFormat.setFilterPredicates(job.getConfiguration, filterResolver)
-      queryModel.setFilterExpressionResolverTree(filterResolver)
+      // before applying filter check whether segments are available in the table.
+      val splits = carbonInputFormat.getSplits(job)
+      if (!splits.isEmpty) {
+        var filterResolver = carbonInputFormat
+          .getResolvedFilter(job.getConfiguration, filterExpression)
+        CarbonInputFormat.setFilterPredicates(job.getConfiguration, filterResolver)
+        queryModel.setFilterExpressionResolverTree(filterResolver)
+      }
     }
     catch {
       case e: Exception =>
--- a/integration/spark/src/test/scala/org/carbondata/spark/testsuite/filterexpr/FilterProcessorTestCase.scala
+++ b/integration/spark/src/test/scala/org/carbondata/spark/testsuite/filterexpr/FilterProcessorTestCase.scala
@@ -19,14 +19,14 @@
 
 package org.carbondata.spark.testsuite.filterexpr
 
-import org.apache.spark.sql.common.util.CarbonHiveContext._
-import org.apache.spark.sql.common.util.QueryTest
-import org.apache.spark.sql.Row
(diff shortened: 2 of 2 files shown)


integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -80,11 +80,14 @@ class CarbonQueryRDD[K, V](
...
     // set filter resolver tree
     try {
-      var filterResolver = carbonInputFormat
-        .getResolvedFilter(job.getConfiguration, filterExpression)
-
-      CarbonInputFormat.setFilterPredicates(job.getConfiguration, filterResolver)
-      queryModel.setFilterExpressionResolverTree(filterResolver)
+      //before applying filter check whether segments are available in the table.
+      val splits = carbonInputFormat.getSplits(job)
+      if (!splits.isEmpty) {
```
manishgupta88: @mohammadshahidkhan ....we have this check after this try catch block...can you please move this code under that check...it will be better and it will avoid same "splits.isEmpty" checks multiple times

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -80,11 +80,14 @@ class CarbonQueryRDD[K, V](
...
     // set filter resolver tree
     try {
-      var filterResolver = carbonInputFormat
-        .getResolvedFilter(job.getConfiguration, filterExpression)
-
-      CarbonInputFormat.setFilterPredicates(job.getConfiguration, filterResolver)
-      queryModel.setFilterExpressionResolverTree(filterResolver)
+      // before applying filter check whether segments are available in the table.
+      val splits = carbonInputFormat.getSplits(job)
+      if (!splits.isEmpty) {
```
manishgupta88: @mohammadshahidkhan ....we have this check after this try catch block...can you please move this code under that check...it will be better and it will avoid same "splits.isEmpty" checks multiple times

kumarvishal09:
I think this issue was fixed earlier also

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -80,11 +80,14 @@ class CarbonQueryRDD[K, V](
     val LOGGER = LogServiceFactory.getLogService(this.getClass.getName)
     // set filter resolver tree
     try {
-      var filterResolver = carbonInputFormat
-        .getResolvedFilter(job.getConfiguration, filterExpression)
-
-      CarbonInputFormat.setFilterPredicates(job.getConfiguration, filterResolver)
-      queryModel.setFilterExpressionResolverTree(filterResolver)
+      // before applying filter check whether segments are available in the table.
+      val splits = carbonInputFormat.getSplits(job)
```
Vimal-Das: Move this outside this try catch block, get the splits only once and have the check only once.

mohammadshahidkhan: Please check the getSplits function if filterpredicate is set then we have different handling. so calling two time is needed.
Code snipet: 

if (filterPredicates == null) {
        return getSplitsInternal(job);
      } else {
        if (filterPredicates instanceof Expression) {
          //process and resolve the expression.
          CarbonInputFormatUtil.processFilterExpression((Expression) filterPredicates, carbonTable);
          return getSplits(job, CarbonInputFormatUtil.resolveFilter((Expression) filterPredicates,
              getAbsoluteTableIdentifier(job.getConfiguration())));
        } else {
          //It means user sets already resolved expression.
          return getSplits(job, (FilterResolverIntf) filterPredicates);
        }
      }

But yes we should avoid two time calling the same if the filter is not applied.

Vimal-Das:
LGTM, Merged to master

PR #762, opened 2016-06-27
Fixed count star with filter issue

kumarvishal09:
When no measure present in the carbon table we are adding default measure . in case of count (*) with filter query by default we are adding default measure but in this case we first measure is a dummy measure which will not have add data so in that case adding first dimension


Vimal-Das:
LGTM, Merged to master

PR #763, opened 2016-06-27
Btree loading performance improvement

kumarvishal09:
When number of block per task is more in that case it was getting loaded one by one and it was slow, now added thread to load the btree , if blocks are already loaded then it will not load ,if it is not loaded it will launch a thread task to load the btree. thread pool size is number of core given assigned for query ... as this is a part of query execution. 

Test cases are already present refer BlockIndexStoreTest.java


gvramana:
LGTM

PR #767, opened 2016-06-27
[WIP][CARBONDATA-16] BLOCK distribution in query is not correct in query when number of executors are less than the cluster size.

mohammadshahidkhan:
Fixed https://issues.apache.org/jira/browse/CARBONDATA-16


kumarvishal09:
can u please provide more detail about the issue...:)

mohammadshahidkhan:
Currently the scenario when no executor is in active  mode, the block distribution will not be possible.
Since the current implementation is expecting the active node.
Before getting the list of active executor node we have to request the required no of executor.

mohammadshahidkhan:
Closing this PR raised similar PR in incubator branch 
https://github.com/apache/incubator-carbondata/pull/20

PR #771, opened 2016-06-29
CARBONDATA-17 - select count(*) from table where column_x = 'value' is not returning the correct count

mohammadshahidkhan:
CARBONDATA-17 - select count(*) from table where column_x = 'value' is not returning the correct count if the first measure is having all values null.

### DIFF
--- a/core/src/main/java/org/carbondata/core/constants/CarbonCommonConstants.java
+++ b/core/src/main/java/org/carbondata/core/constants/CarbonCommonConstants.java
@@ -155,6 +155,10 @@ public final class CarbonCommonConstants {
    */
   public static final String COUNT = "count";
   /**
+   * Count constant
+   */
+  public static final String COUNT_STAR = "countstar";
+  /**
    * Max constant
    */
   public static final String MAX = "max";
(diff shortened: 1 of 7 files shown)


gvramana:
retest this please

kumarvishal09:
retest this please

processing/src/main/java/org/carbondata/processing/graphgenerator/GraphGenerator.java
```diff
@@ -472,7 +472,8 @@ private StepMeta getSliceMeregerStep(GraphConfigurationInfo configurationInfo,
       String[] aggType = configurationInfo.getAggType();
       StringBuilder builder = new StringBuilder();
       for (int i = 0; i < aggType.length - 1; i++) {
-        if (aggType[i].equals(CarbonCommonConstants.COUNT)) {
```
kumarvishal09: no need to change here as this is data loading flow and count star with filter for query execution

kumarvishal09:
LGTM , except one small comment.

processing/src/main/java/org/carbondata/processing/graphgenerator/GraphGenerator.java
mohammadshahidkhan: fixed

PR #774, opened 2016-06-29
[BUG] Min Max select query on decimal columns are failing with ClassCastException in Spark beeline

nareshpr:


integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -576,7 +614,14 @@ case class MinFunctionCarbon(expr: Expression, base: AggregateExpression1, final
       if (min.isFirstTime) {
         null
       } else {
-        Cast(Literal(min.getValueObject), base.dataType).eval(null)
+        min match {
+          case s: MaxBigDecimalAggregator =>
```
gvramana: why min matches maxaggreagator

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -576,7 +614,14 @@ case class MinFunctionCarbon(expr: Expression, base: AggregateExpression1, final
       if (min.isFirstTime) {
         null
       } else {
-        Cast(Literal(min.getValueObject), base.dataType).eval(null)
+        min match {
+          case s: MinBigDecimalAggregator =>
```
gvramana: add test cases covering datatypes min,max cases

nareshpr: Fixed

nareshpr:
retest this please

PR #775, opened 2016-06-29
ColumnGroup[Exclude filter and RowLevel filter]

ashokblend:
Fix for Exclude and Rowlevel filter
Handled issue https://issues.apache.org/jira/browse/CARBONDATA-20


core/src/main/java/org/carbondata/query/filter/executer/RowLevelFilterExecuterImpl.java
```diff
@@ -144,18 +150,20 @@ private void createRow(BlocksChunkHolder blockChunkHolder, RowIntf row, int inde
     Object[] record = new Object[dimColEvaluatorInfoList.size() + msrColEvalutorInfoList.size()];
     String memberString = null;
     for (DimColumnResolvedFilterInfo dimColumnEvaluatorInfo : dimColEvaluatorInfoList) {
+      int blockIndex = segmentProperties.getDimensionOrdinalToBlockMapping()
```
gvramana: check if column group column, then only do map lookup

ashokblend: added array to lookup blockindex

Issue #777, opened 2016-06-29
Filter query issue for >, <, <= than filter


mohammadshahidkhan:
#746 Filter query issue for >, <, <= than filter

1. select count(empid) from a12 where dob > '2014-07-01 12:07:28'
   throwing runtime exception
2. select count(empid) from a12 where dob < '2014-07-01 12:07:28'
   is including the null values also.
3. select count(empid) from a12 where dob <=  '2014-07-01 12:07:28'
   is including the null value 
Null should not considered in less than filter.
4. select count(empid) from a12 where dob <=  '2014-07-01 12:07:28' is not returning proper value.
5.select count(empid) from a12 where dob >=  '2014-07-01 12:07:28' is not returning proper value.

create cube command: 
create table a12(empid String,ename String,sal double,deptno int,mgr string,gender string," +
        "dob timestamp,comm decimal(4,2),desc string) stored by 'org.apache.carbondata.format
data:
empid,ename,sal,deptno,mgr,gender,dob,comm,desc
1,abc,1233,10,2,,2014-07-01 12:07:28,1234.191,string_null
2,bcd,1322,,3,f,2014-07-01 12:07:28,19.99,int_null
3,cde,4322,,4,m,,16.996,date_null
4,    ,43243,,5,m,,999.117,string_space
5,,43242,20,6,m,2017-07-01 12:07:28,99.999,string_null
6,ijk,,20,6,m,2017-07-01 12:07:28,50089,double_null
7,pqr,2422,20,6,m,2017-07-01 12:07:28,32.339,decimal_null
8

PR #778, opened 2016-06-29
[issue-777] Filter query issue for >, <, <= than filter

mohammadshahidkhan:
Fixed: #777


core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
```diff
@@ -153,6 +169,79 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+    BitSet bitSet = new BitSet(numerOfRows);
+    int[] columnIndex = dimensionColumnDataChunk.getAttributes().getInvertedIndexes();
+    int start = 0;
+    int last = 0;
+    int startIndex = 0;
+    int skip = 0;
+    byte[][] filterValues = this.filterRangeValues;
+
+    //find the number of default values to skip the null value in case of direct dictionary
+    for (int i = 0; i < numerOfRows; i++) {
```
kumarvishal09: all the blocks are sorted so all the default value will  be at first ..so we we can assign the startIndext=skip so our actual search can start from that point

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
```diff
@@ -153,6 +169,79 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+          .getFirstIndexUsingBinarySearch(dimensionColumnDataChunk, startIndex, numerOfRows - 1,
+              filterValues[i]);
+      // Logic will handle the case where the range filter member is not present in block
+      // in this case the binary search will return the index from where the bit sets will be
+      // set inorder to apply filters. this is Lesser than filter so the range will be taken
+      // from the prev element which is Lesser than filter member.
+      start = CarbonUtil.nextLesserValueToTarget(start, dimensionColumnDataChunk, filterValues[i]);
+      if (start < 0) {
+        start = -(start + 1);
+        if (start == numerOfRows) {
```
kumarvishal09: i think in both if and else if condition we are doing the same thing one if check with >= condition will work fine

core/src/main/java/org/carbondata/query/filters/measurefilter/util/FilterUtil.java
```diff
@@ -587,11 +587,29 @@ public static DimColumnFilterInfo getFilterListForAllMembersRS(Expression expres
           LOGGER.error(e.getMessage());
         }
       }
+
     }
     return filterValuesList.toArray(new byte[filterValuesList.size()][]);
 
   }
 
+  public static byte[] getMaskKey(int surrogate, CarbonDimension carbonDimension,
```
kumarvishal09: duplicate method,  method to get the filter value is already present,

mohammadshahidkhan: this method is to get the value of single surrogate key mask value,  not  for dimension filter.

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanEqualFilterExecuterImpl.java
```diff
@@ -148,6 +163,72 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+    int start = 0;
+    int last = 0;
+    int startIndex = 0;
+    byte[][] filterValues = this.filterRangeValues;
+    int skip = 0;
+    //find the number of default values to skip the null value in case of direct dictionary
+    for (int i = 0; i < numerOfRows; i++) {
+      int comp = ByteUtil.UnsafeComparer.INSTANCE
+          .compareTo(dimensionColumnDataChunk.getCompleteDataChunk(),
+              skip * defaultValue.length, defaultValue.length, defaultValue, 0,
```
gvramana: if nulls are more, it becomes row by row comparision, instead do binary search with value+1 so that next matching value

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
```diff
@@ -153,6 +169,79 @@ private BitSet setFilterdIndexToBitSetWithColumnIndex(
...
+   * will be considered for applying range filters. this method will be called if the
+   * column is not supported by default so column index mapping  will be present for
+   * accesing the members from the block.
+   *
+   * @param dimensionColumnDataChunk
+   * @param numerOfRows
+   * @return BitSet.
+   */
+  private BitSet setFilterdIndexToBitSetWithColumnIndexForDirectDictionary(
+      FixedLengthDimensionDataChunk dimensionColumnDataChunk, int numerOfRows,
```
gvramana: modify existing function, not required to copy paste code

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
mohammadshahidkhan: fixed

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanEqualFilterExecuterImpl.java
mohammadshahidkhan: fixed

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
mohammadshahidkhan: fixed

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanEqualFilterExecuterImpl.java
```diff
@@ -109,13 +124,26 @@ private BitSet getFilteredIndexes(DimensionColumnDataChunk dimensionColumnDataCh
...
     BitSet bitSet = new BitSet(numerOfRows);
     int[] columnIndex = dimensionColumnDataChunk.getAttributes().getInvertedIndexes();
     int start = 0;
     int last = 0;
+    int skip = 0;
     int startIndex = 0;
     byte[][] filterValues = this.filterRangeValues;
+    //find the number of default values to skip the null value in case of direct dictionary
+    if (null != defaultValue) {
+      start = CarbonUtil
```
kumarvishal09: Please correct me if I am wrong what about multiple null value in one block binary search will give first position, I think we need to get the last position.

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanEqualFilterExecuterImpl.java
```diff
@@ -109,13 +124,26 @@ private BitSet getFilteredIndexes(DimensionColumnDataChunk dimensionColumnDataCh
...
+          .getFirstIndexUsingBinarySearch(dimensionColumnDataChunk, startIndex, numerOfRows - 1,
+              defaultValue);
+      if (start < 0) {
```
kumarvishal09: This condition is required? if null value is not present then skip will be 0

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
```diff
@@ -109,13 +125,28 @@ private BitSet getFilteredIndexes(DimensionColumnDataChunk dimensionColumnDataCh
...
     int[] columnIndex = dimensionColumnDataChunk.getAttributes().getInvertedIndexes();
     int start = 0;
     int last = 0;
     int startIndex = 0;
+    int skip = 0;
     byte[][] filterValues = this.filterRangeValues;
+
+    //find the number of default values to skip the null value in case of direct dictionary
+    if (null != defaultValue) {
+      start = CarbonUtil
```
kumarvishal09: Same as above comment

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanFiterExecuterImpl.java
```diff
@@ -109,13 +125,28 @@ private BitSet getFilteredIndexes(DimensionColumnDataChunk dimensionColumnDataCh
...
+          .getFirstIndexUsingBinarySearch(dimensionColumnDataChunk, startIndex, numerOfRows - 1,
+              defaultValue);
+      if (start < 0) {
```
kumarvishal09: Same as above

core/src/main/java/org/carbondata/query/filter/executer/RowLevelRangeLessThanEqualFilterExecuterImpl.java
mohammadshahidkhan: @kumarvishal09 here we are getting the first non null value index in the dimensionColumnDataChunk, because in default we are not passing the null value surrogate but the defaultSurrogate+ 1

please look back getFilteredIndexes 
 int key = directDictionaryGenerator.generateDirectSurrogateKey(null) + 1;
        byte[] defaultValue = FilterUtil
            .getMaskKey(key, dimColEvaluatorInfoList.get(0).getDimension(),
                this.segmentProperties.getDimensionKeyGenerator());
        return setFilterdIndexToBitSetWithColumnIndex(
            (FixedLengthDimensionDataChunk) dimensionColumnDataChunk, numerOfRows, defaultValue);

gvramana:
LGTM, reviewed

PR #779, opened 2016-06-29
[BUG] Timestamp column filter is not working

nareshpr:


core/src/main/java/org/carbondata/query/filters/FilterExpressionProcessor.java
```diff
@@ -333,18 +334,10 @@ private FilterResolverIntf getFilterResolverBasedOnExpressionType(
...
-              } else if (expressionTree.getFilterExpressionType() == ExpressionType.UNKNOWN) {
-                return new RowLevelFilterResolverImpl(expression, false, false, tableIdentifier);
-              }
-
-              return new ConditionalFilterResolverImpl(expression, true, true);
-            }
+          if (condExpression.getColumnList().get(0).getCarbonColumn()
+              .hasEncoding(Encoding.DICTIONARY) && !condExpression.getColumnList().get(0)
+              .getCarbonColumn().hasEncoding(Encoding.DIRECT_DICTIONARY)) {
+            return new ConditionalFilterResolverImpl(expression, true, true);
```
gvramana: add else case to fall back to row level resolver

nareshpr: Fixed

PR #787, opened 2016-07-05
[CARBONDATA-16] BLOCK distribution in query is not correct in query when number of executors are less than the cluster size.

mohammadshahidkhan:
Fixed : https://issues.apache.org/jira/browse/CARBONDATA-16


integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -107,8 +108,25 @@ class CarbonQueryRDD[K, V](
       )
       if (blockList.nonEmpty) {
         // group blocks to nodes, tasks
+        var activeNodes = DistributionUtil.getNodeList(sparkContext)
```
gvramana: use nodelist only once.
nodeBlockTaskMapping also should be used only once

mohammadshahidkhan: Fixed

mohammadshahidkhan:
fixed review comments
Please review

gvramana:
retest this please

PR #791, opened 2016-07-07
[Bug] Changed the way Subquery matching is done

nareshpr:


integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
```diff
@@ -142,7 +142,7 @@ private[sql] case class CarbonDatasourceRelation(
                     carbonRelation.cubeMeta.carbonTableIdentifier.getDatabaseName +
                     CarbonCommonConstants.FILE_SEPARATOR +
                     carbonRelation.cubeMeta.carbonTableIdentifier.getTableName
-    val fileType = FileFactory.getFileType
+    val fileType = FileFactory.getFileType(tablePath)
```
gvramana: Please do data load and check if it refreshes size, every time

PR #792, opened 2016-07-07
[Bug] Broadcast join is not reloading the size after data loading

nareshpr:


eason-lyx:
@nareshpr the sizeInBytes update is no working except restart thrift server.

PR #796, opened 2016-07-12
[Bug] Handling the dynamic executor allocation during compaction.

ravikiran23:
for dynamic executor allocation, taking the max executors parameter and requesting the required executors.


gvramana:
LGTM

PR #797, opened 2016-07-12
[Bug] Drop Table dbname.tablename is not working, carbondata file path if '-' exist, gives wrong value.

nareshpr:


ravipesala:
LGTM

PR #800, opened 2016-07-13
inverted index configurable

Zhangshunyu:
Using DDL to determine which column not do inverted index, if the user doesn't specify for one column, the default value is true which means do inverted index.
For example:
```

(id Int, name String, city String)
TBLPROPERTIES('NO_INVERTED_INDEX'='name,city')
```


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -649,6 +653,45 @@ class CarbonSqlParser()
...
   /**
+   * This will extract the no inverted columns fields.
+   * By default all dimensions use inverted index.
+   *
+   * @param fields
+   * @param tableProperties
+   * @return
+   */
+  protected def extractNoInvertedIndexColumns(fields: Seq[Field],
+                                              tableProperties: Map[String, String]):
```
QiangCai: code style

processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -1238,28 +1252,37 @@ private Consumer(BlockletDataHolder blockletDataHolder) {
     private byte[][] data;
     private boolean isSortRequired;
     private boolean isCompressionReq;
+    private boolean isUseInvertedIndex;
 
     private boolean isNoDictionary;
 
-    private BlockSortThread(int index, byte[][] data, boolean isSortRequired) {
+    private BlockSortThread(int index, byte[][] data, boolean isSortRequired,
+                            boolean isUseInvertedIndex) {
```
QiangCai: code style

core/src/main/java/org/carbondata/core/datastorage/store/columnar/BlockIndexerStorageForNoInvertedIndex.java
```diff
@@ -0,0 +1,159 @@
...
+import org.carbondata.core.util.ByteUtil;
+
+public class BlockIndexerStorageForNoInvertedIndex implements IndexStorage<int[]> {
+  private byte[][] keyBlock;
+  private byte[][] sortedBlock;
+  private int totalSize;
+  private int[] dataIndexMap;
+
+  public BlockIndexerStorageForNoInvertedIndex(byte[][] keyBlockInput, boolean compressData,
+                                               boolean isNoDictionary) {
```
QiangCai: code style

QiangCai:
LGTM

PR #801, opened 2016-07-13
[CARBONDATA-35] generate global dict using pre-defined dict from external column file

Jay357089:
To support pre-defined dict from external column file, we add Key word COLUMNDICT in option of load DML to define the column name and column file path as below:

LOAD DATA inpath 'D:/inputPath/carbonTest.csv' INTO TABLE carbonTest options( 'COLUMNDICT '='name:D:/Carbon/carbonData/name.csv,sex:D:/Carbon/carbonData/sex.csv'), 
user needs to set as "column_Name: column_Path", and multiple columns and file paths can be delimited by ",'.
for complex type, user needs to set directly on primitive type. For example, for struct type(imei struct(num int, name string)), user needs set like "imei.num:columfilepath", for Array type(imei Array(string)), user can set "imei:columfilepath". for more , pls check ExternalColumnDictionaryTestCase.scala

we handle the above DML as below:
1. in LoadCube, set column name and dictionanry file path in carbonLoadModel.
2. get the column name and column file path from carbonLoadModel, test whether the column is exists, then generate new dict using the column file.
3. read the old dictionary file and compare with new dict and remove duplication, then write dict file.
4 if the column data is also provided by fact table, then drop the column data from fact table when generating dict using fact table file.


Jay357089:
@chenliang613 @QiangCai pls review.

eason-lyx:
retest this please

eason-lyx:
LGTM

PR #802, opened 2016-07-13
CARBONDATA-5 data mismatch between the carbon Table and Hive Table for columns having \N for non numeric data type

mohammadshahidkhan:
data mismatch between the carbon Table and Hive Table for columns having \N for non numeric data type
Modification: 
A property "serialization.null.format" is exposed to configure the value to be treated as null.
Default value of key serialization.null.format to be treated as null is \N.


mohammadshahidkhan:
image

conf/carbon.properties.template
```diff
@@ -95,4 +95,6 @@ carbon.enable.quick.filter=false
 ##The property to set the date to be considered as start date for calculating the timestamp.
 #carbon.cutOffTimestamp=2000-01-01 00:00:00
 ##The property to set the timestamp (ie milis) conversion to the SECOND, MINUTE, HOUR or DAY level.
-#carbon.timegranularity=SECOND
\ No newline at end of file
+#carbon.timegranularity=SECOND
```
sujith71955: i think carbon.timegranularity change showing is not proper, pls correct format so that git wont show this change

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -81,6 +81,7 @@ class CarbonSqlParser()
   protected val FACT = Keyword("FACT")
   protected val FIELDS = Keyword("FIELDS")
   protected val FILEHEADER = Keyword("FILEHEADER")
+  protected val SERIALIZATION_NULL_FORMAT = Keyword("SERIALIZATION.NULL.FORMAT")
```
sujith71955: Follow other Keyword format, better to use underscores rather than dot.

processing/src/main/java/org/carbondata/processing/graphgenerator/GraphGenerator.java
```diff
@@ -794,6 +796,7 @@ private GraphConfigurationInfo getGraphConfigInfoForFact(
     prepareNoDictionaryMapping(dimensions, graphConfiguration);
     graphConfiguration
         .setColumnSchemaDetails(CarbonSchemaParser.getColumnSchemaDetails(dimensions));
+    graphConfiguration.setTableOptionWrapper(prepareTableOptionWrapper());
```
sujith71955: i think getTableOptionWrapper is more readable  since as per logic its just getting the format.

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -874,7 +880,7 @@ private String getCarbonLocalBaseStoreLocation() {
     int i = 0;
     for (Object obj : rowValue) {
       if (obj != null) {
-        if (obj.equals(valueToCheckAgainst)) {
+        if (obj.equals(serializationNullFormat) || obj.equals(valueToCheckAgainst)) {
```
sujith71955: better to have equalsignore case, some time user can also mention as \N-\n

processing/src/main/java/org/carbondata/processing/schema/metadata/TableOptionWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * The class hold the table option details being used while dataload
+ */
+public class TableOptionWrapper {
+  private Map<String, TableOption> mapOFOptions = null;
+
+  /**
+   * @param input
+   */
+  public TableOptionWrapper(String input) {
```
sujith71955: while creating each table option do we need to create a  map? should be singleton i guess. so that the map can hold all table options

processing/src/main/java/org/carbondata/processing/schema/metadata/TableOptionWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+    mapOFOptions = new HashMap<>(CarbonCommonConstants.DEFAULT_COLLECTION_SIZE);
+    String[] split =
+        null != input ? input.split(CarbonCommonConstants.HASH_SPC_CHARACTER) : new String[0];
```
sujith71955: no where i can see where the input string is appended with HASH_SPC_CHAR so that it can be split, can you please show where its been added

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
mohammadshahidkhan: \n we can not expect here since its a new line charector

conf/carbon.properties.template
mohammadshahidkhan: fixed

processing/src/main/java/org/carbondata/processing/schema/metadata/TableOptionWrapper.java
mohammadshahidkhan: currently we are adding only single property in the wrapper ie why not seen,
When we will pass pass some more property then it will be used.

processing/src/main/java/org/carbondata/processing/graphgenerator/GraphGenerator.java
mohammadshahidkhan: fixed

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
mohammadshahidkhan: fixed

processing/src/main/java/org/carbondata/processing/schema/metadata/TableOptionWrapper.java
mohammadshahidkhan: fixed

sujith71955:
retest this please

PR #803, opened 2016-07-13
[CARBONDATA-53][BUG]Handle query result null for negative number in No_dictionary Column

Jay357089:
https://issues.apache.org/jira/browse/CARBONDATA-53

the original code uses 'isDigit' to check number, while negative number has ‘-’， which is not digitm, then return null. Change to use "isNumber"


Jay357089:
@ravipesala  pls review.

kumarvishal09:
LGTM

kumarvishal09:
retest this please

PR #804, opened 2016-07-13
[CARBONDATA-36]fix dictionary exception when data column count less then schema

eason-lyx:
when csv head is id,name,age,city
if the data record like as the following:
1,a,10,china
2,b
3,c,11,usa
then,when read this row:2,b,generating dicionary will failure,and the remain record (3,c,11,usa)will no generate dictionary.

this pr to prevent put null value to Set,and let it continue.

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
@@ -73,7 +73,7 @@ case class PrimitiveParser(dimension: CarbonDimension,
 
   def parseString(input: String): Unit = {
     if (hasDictEncoding) {
-      if (StringUtils.isNotEmpty(input)) {
+      if (input != null) {
         set.add(input)
       }
     }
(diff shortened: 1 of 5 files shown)


integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -67,7 +67,9 @@ case class PrimitiveParser(dimension: CarbonDimension,
 
   def parseString(input: String): Unit = {
     if (hasDictEncoding) {
-      set.add(input)
+      if (StringUtils.isNotEmpty(input)) {
```
kumarvishal09: empty value in data is a valid scenario , here as we are using hashtable we cannot insert null value i think null check is ok..Please correct me if i am worng

eason-lyx: yes empty value is a valid scenario.
in dictionary file,we already contain the null value.so here we no need to insert to set,also set no allow insert null.so dictionary generate fail in this place.
and in csv reader:we get the row like:1,abc,null,null,null
we can match the null value in dictionary file.

kumarvishal09: please update the code

Vimal-Das:
LGTM

Vimal-Das:
Merged to master.

PR #807, opened 2016-07-14
[CARBONDATA-52] fix query average value of bigInt filed

foryou2030:
when query average value of bigInt filed "column1", 
if the value of "sum(column1)/count(column1)" is double type with decimal,
the result of avg(column1) is inaccurate, it lost precision.

this pr use double type value as avg result, fix the issue


ravipesala:
LGTM

PR #808, opened 2016-07-14
[Bug]Columngroup with AutoDetect Highcardinality

ashokblend:
if it is column group column or dictionary included column than do not check for high cardinality.


Vimal-Das:
retest this please

Vimal-Das:
LGTM, Merged to master.

PR #809, opened 2016-07-14
CARBONDATA-56 Exception thrown when aggregation on dimension which has decimal datatype

manishgupta88:
Problem: In case of aggregation on dimension which has decimal datatype, the data from utility class is returned as spark sql decimal type. Because of this while aggregating this data data with java decimal type class cast exception is thrown.

Fix: Type cast to java decimal type while aggregating dimension data for decimal type


Vimal-Das:
LGTM, merged to master.

PR #810, opened 2016-07-14
[CARBONDATA-57] BLOCK distribution un wanted wait for the executor node even though the sufficient nodes are available

mohammadshahidkhan:
Modification:
No need to wait if the existing executors are more than or equal to the required executors.


gvramana:
retest this please

PR #811, opened 2016-07-14
[Bug] Describe formatted on hive table will be pushed to spark and result will be shown as per spark output

nareshpr:


Vimal-Das:
LGTM, Merged to Master

PR #812, opened 2016-07-14
[Bug] Dataload is failing by flipping datatype of string column to array<string> column

nareshpr:


ravipesala:
LGTM. Please rebase the code.

PR #815, opened 2016-07-14
[Bug]Column group count query

ashokblend:
This change fixed Count query on column group.


Vimal-Das:
LGTM, Merged to master.

PR #816, opened 2016-07-14
[BUG] Invalid column name validation in column groups.

ravikiran23:
if any invalid column is given in the column group then create should fail with error message.


Vimal-Das:
LGTM, Merged to master.

PR #817, opened 2016-07-14
[Bug] For fixing arithmetic expression filter issues in Filters.

sujith71955:
a) Issue was there since the filter expression with arithmetic expression with no dictionary column was not evaluating Rowwise, since the system has to resolve first the add expression system has to resolve the filter using RowLevelResolver.


Vimal-Das:
retest this please

Vimal-Das:
LGTM, Merged to master.

PR #818, opened 2016-07-14
[Bug] Order by with limit in case of non aggregation query shall not …

sujith71955:
Order by with limit in case of non aggregation query shall not be pushed down to carbon layer, since  it was pushing down queries were failing which consists of order by limit


gvramana:
LGTM

PR #819, opened 2016-07-15
CARBONDATA-59 Filter queries on columns other than string datatype cannot get the correct result when included as dictionary column

manishgupta88:
Problem: when a column with datatype other than string is added to dictionary include property, then filters queries on that column cannot get the correct result

Analysis: In Sort index file always the numeric values comes first and then the string values. When we perform binary search then when a string value is encountered then parsing operation to correspond datatype throws an exception and -1 value is returned because of which we move further towards last index but actually we should move towards lower index as all numeric values are above the string values.

Fix: return 1 in case of parsing operation failure because of which we will move towards lower index in binary search operation.


Vimal-Das:
LGTM, Merged to master.

PR #821, opened 2016-07-15
[BUG] Complex type decimal support

ravikiran23:
Decimal type was failing in the complex type. 
Parsing of the decimal data type was wrong , corrected the same.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -300,7 +300,7 @@ class CarbonSqlParser()
     m.find()
     val matchedString: String = m.group(1)
     val scaleAndPrecision = matchedString.split(",")
-    (Integer.parseInt(scaleAndPrecision(0)), Integer.parseInt(scaleAndPrecision(1)))
+    (Integer.parseInt(scaleAndPrecision(0).trim), Integer.parseInt(scaleAndPrecision(1).trim))
   }
 
   /**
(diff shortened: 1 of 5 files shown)


integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -1147,30 +1162,55 @@ class CarbonSqlParser()
...
-    field.dataType.getOrElse("NIL") match {
+    val dataType = field.dataType.getOrElse("NIL")
+    dataType match {
       case "string" => Field(field.column, Some("String"), field.name, Some(null), field.parent,
-        field.storeType)
+        field.storeType
+      )
       case "integer" | "int" => Field(field.column, Some("Integer"), field.name, Some(null),
-        field.parent, field.storeType)
+        field.parent, field.storeType
```
sujith71955: format is not proper, displaying non modified code

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -300,7 +300,7 @@ class CarbonSqlParser()
     m.find()
     val matchedString: String = m.group(1)
     val scaleAndPrecision = matchedString.split(",")
-    (Integer.parseInt(scaleAndPrecision(0)), Integer.parseInt(scaleAndPrecision(1)))
+    (Integer.parseInt(scaleAndPrecision(0).trim), Integer.parseInt(scaleAndPrecision(1).trim))
```
sujith71955: Add test cases, if already present please mention the test class

ravikiran23: added

sujith71955:
LGTM

PR #822, opened 2016-07-15
[WIP][Bug]Handling for percentile aggregate function

ashokblend:
We convert int to double while data loading or query. We need to change it to BigInt

### DIFF
--- a/core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
+++ b/core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
@@ -33,6 +33,8 @@ public final class DataTypeUtil {
         BigDecimal bigDecimal =
             new BigDecimal(msrValue).setScale(carbonMeasure.getScale(), RoundingMode.HALF_UP);
         return normalizeDecimalValue(bigDecimal, carbonMeasure.getPrecision());
+      case INT:
+        return Double.valueOf(msrValue).longValue();
       case LONG:
         return Long.valueOf(msrValue);
       default:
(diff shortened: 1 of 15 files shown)


ashokblend:
retest this please

ashokblend:
retest this please

ashokblend:
retest this pleaswe

ashokblend:
retest this please

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -33,6 +33,8 @@ public static Object getMeasureValueBasedOnDataType(String msrValue, DataType da
         BigDecimal bigDecimal =
             new BigDecimal(msrValue).setScale(carbonMeasure.getScale(), RoundingMode.HALF_UP);
         return normalizeDecimalValue(bigDecimal, carbonMeasure.getPrecision());
+      case INT:
+        return Double.valueOf(msrValue).longValue();
```
ravipesala: Why not `Long.valueOf(msrValue)` ?

ravipesala:
LGTM except one comment.

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
ashokblend: If user gives double value in csv, instead of int.  Thats why i have added Double.valueOf(msrValue).longValue().
It hold good for Long also. Need to check how to handle this scenario.

In this scenario, Hive result the integer value but carbon return null.

ravipesala: ok

PR #824, opened 2016-07-16
[Bug] Filter was failing when join condition

sujith71955:
Filter was failing when join condition is been applied between two tables,spark will send optimized logical plan for execution where
timestamp type will change to Long and this since the left expression and right expression was getting different data type issue was happening in case of Timestamp type column filters, For other filter like BigInt and Decimal, expression result was not handling such data type while comparing the expression results.

### DIFF
--- a/core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
+++ b/core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
@@ -365,12 +365,15 @@ public class ExpressionResult implements Comparable<ExpressionResult> {
         case IntegerType:
           result = this.getInt().equals(objToCompare.getInt());
           break;
-
+        case LongType:
+        case TimestampType:
+          result = this.getLong().equals(objToCompare.getLong());
+          break;
         case DoubleType:
           result = this.getDouble().equals(objToCompare.getDouble());
           break;
-        case TimestampType:
-          result = this.getLong().equals(objToCompare.getLong());
+        case DecimalType:
+          result = this.getDecimal().equals(objToCompare.getDecimal());
           break;
         default:
           break;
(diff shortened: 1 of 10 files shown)


sujith71955:
retest this please

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
```diff
@@ -365,12 +365,15 @@ public Boolean getBoolean() throws FilterIllegalMemberException {
...
+        case TimestampType:
+          result = this.getLong().equals(objToCompare.getLong());
+          break;
         case DoubleType:
           result = this.getDouble().equals(objToCompare.getDouble());
           break;
-        case TimestampType:
-          result = this.getLong().equals(objToCompare.getLong());
+        case DecimalType:
+          result = this.getDecimal().equals(objToCompare.getDecimal());
```
sraghunandan: java suggests to use compareTo for comparing values

core/src/main/java/org/carbondata/query/expression/ExpressionResult.java
```diff
@@ -365,12 +365,15 @@ public Boolean getBoolean() throws FilterIllegalMemberException {
         case IntegerType:
           result = this.getInt().equals(objToCompare.getInt());
           break;
-
+        case LongType:
+        case TimestampType:
+          result = this.getLong().equals(objToCompare.getLong());
+          break;
         case DoubleType:
           result = this.getDouble().equals(objToCompare.getDouble());
```
sraghunandan: should use compareTo to compare values

sraghunandan: should use compare for comparing values

core/src/main/java/org/carbondata/query/expression/conditional/InExpression.java
```diff
@@ -48,40 +48,34 @@ public InExpression(Expression left, Expression right) {
...
+            val = new ExpressionResult(val.getDataType(), expressionResVal.getString());
+            break;
+          case IntegerType:
+            val = new ExpressionResult(val.getDataType(), expressionResVal.getInt());
+            break;
+          case DoubleType:
+            val = new ExpressionResult(val.getDataType(), expressionResVal.getDouble());
+            break;
+          case LongType:
+          case TimestampType:
```
gvramana: Only comparision in ExpressionResult should take care that you get long and compare, expressions should not return long type.

Zhangshunyu:
retest this please

sujith71955:
retest this please

PR #830, opened 2016-07-18
[CARBONDATA-62] Values not valid for column datatype are not getting discarded while generating global dictionary

manishgupta88:
If value read from raw data is not valid for its datatype then discard that value at the time of data loading and insert null at its place while storing data in carbon format. 
If not done so, filter queries other than isNull condition will display all the null values in the result


manishgupta88:
restest this please

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -189,4 +189,39 @@ public static Object getDataBasedOnDataType(String data, DataType actualDataType
...
   }
+
+  /**
+   * This method will parse a given string value corresponding to its datatype
+   *
+   * @param value    value to parse
+   * @param dataType datatype for that value
+   * @return
+   */
+  public static String validateColumnValueForItsDataType(String value, DataType dataType) {
```
gvramana: Change the return type to boolean

PR #827, opened 2016-07-18
[CARBONDATA-67][BUG] key word should not be allowed in database or table name.

ravikiran23:
If a key word is given as the database name or table name then throwing exception in create table flow.

PR #833, opened 2016-07-18
[CARBONDATA-67] removing key words from carbon parser.

ravikiran23:
removing the keyword declaration and converting to regex so that keywords can be used as identifiers.


ravikiran23:
old PR is #827.

PR #837, opened 2016-07-18
[CARBONDATA-76][Bug] Null members should not display in NotEquals Filter

sujith71955:
if the user was applying filte to list down non null members. When user applies Not Equals filter in any non null members the system shall not display null members in report as per Hive compatibility.


core/src/main/java/org/carbondata/query/expression/conditional/NotEqualsExpression.java
```diff
@@ -43,8 +43,11 @@ public NotEqualsExpression(Expression left, Expression right) {
     boolean result = false;
     ExpressionResult val1 = elRes;
     ExpressionResult val2 = erRes;
-
     if (elRes.isNull() || erRes.isNull()) {
+      if (elRes.isNull() && !erRes.isNull()) {
```
gvramana: if any of the LHS or RHS is null , always return false

PR #843, opened 2016-07-19
[CARBONDATA-97][Bug] Decimal Precision and scale is not getting applied based on schema metada

sujith71955:
In case of decimal columns, the decimal values where getting round-off values and in case of 'IN' Filters the result was not coming because of this issue.

In create command the user can set the precision and scale of a decimal column but while querying always the default value is been taken because of which the applied filter values where getting updated resulting in wrong results.


gvramana:
LGTM

PR #845, opened 2016-07-20
[CARBONDATA-86] Value displayed as Null after increase in precision for decimal datatype after aggregation

manishgupta88:
Problem: While creating a table if user specifies a precision for decimal datatype column and then executes an aggregation query on that column then in case if after aggregation the precision of the resulted value becomes greater than the user configure precision value, spark displays the aggregated value as null

Fix: Set the precision of the aggregated result as the precision for the decimal type while sending the aggregated result to spark


integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -366,7 +366,15 @@ case class AverageFunctionCarbon(expr: Expression, base: AggregateExpression1, f
       } else {
         avg match {
           case avg: AvgBigDecimalAggregator =>
-            Cast(Literal(avg.getBigDecimalValue), base.dataType).eval(null)
+            var dataType: DataType = base.dataType
+            val aggResult = avg.getBigDecimalValue
```
sujith71955: move the common code for getting updated decimal precision to a common utility method so that the other case classes can use

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -680,7 +696,20 @@ case class SumDisctinctFunctionCarbon(expr: Expression, base: AggregateExpressio
         null
       }
       else {
-      Cast(Literal(distinct.getValueObject), base.dataType).eval(null)
+        val dataType = base.dataType match {
```
sujith71955: Add testcase for the scenario, if already present please mention the testcase class and id

sujith71955:
LGTM

PR #846, opened 2016-07-20
[CARBONDATA-87]deleting the temp files in case of success / failure of load.

ravikiran23:
deleting the temp files in case of success / failure of load.
deleting the ktr file along with other files.


sujith71955:
retest this please

PR #847, opened 2016-07-21
[Bug] Fix ColumnIdentifier to string to let dictionary generate error information understandable.

Zhangshunyu:
Fix ColumnIdentifier to string to let dictionary generate error information understandable.


Zhangshunyu:
@QiangCai pls review this code, thank you!

core/src/main/java/org/carbondata/core/carbon/ColumnIdentifier.java
```diff
@@ -106,4 +106,8 @@ public DataType getDataType() {
     return true;
   }
 
+  @Override public String toString() {
+    return "ColumnIdentifier{" + "columnId='" + columnId + '\'' + '}';
```
QiangCai: suggest to use '[]' instead of '{}'

Zhangshunyu:
@QiangCai Done, please check again.

kumarvishal09:
@Zhangshunyu can u please add more detail about the issue.I think adding column name will be more useful as columnid is random string

core/src/main/java/org/carbondata/core/carbon/ColumnIdentifier.java
```diff
@@ -106,4 +106,8 @@ public DataType getDataType() {
     return true;
   }
 
+  @Override public String toString() {
+    return "ColumnIdentifier[" + "columnId='" + columnId + '\'' + ']';
```
QiangCai: return "ColumnIdentifier [columnId=" + columnId + "]";

Zhangshunyu:
because there is no column name in ColumnIdentifier and we can see the column id and column name in shcema file. But if we modify the  ColumnIdentifier  class only for this log info, it will affect many places, so we think it is no need to add the column name in ColumnIdentifier  since we can recognize it from the schema file.

kumarvishal09:
Ok. But logging column id wont be useful as user doesn't know about the column id and for the developers it wont be simple to sync columnid with the column name. Please correct me if I am worng

QiangCai:
@kumarvishal09  
We can open schema file directly to find column id and column name. It is easy to map column id with  column name.
After checked code, We found that adding column name for ColumnIdentifier is more difficult than adding table name for CarbonTableIdentifier. I think We need your help.

kumarvishal09:
Ok @QiangCai merge this PR.

QiangCai:
LGTM

PR #848, opened 2016-07-21
[CARBONDATA-90]Struct of array is failing because of complex dimension ordinal

kumarvishal09:
1. Assignment of complex dimension ordinal was not correct as it was not increment and two children was getting same complex dimension ordinal  
2. Second issue was selecting all the block indexes of complex dimension children.


nareshpr:
LGTM, Please add testcase

kumarvishal09:
@nareshpr added testcases please check

sujith71955:
retest this please

sujith71955:
LGTM

PR #850, opened 2016-07-21
[CARBONDATA-93] Task not re-submitted by spark on data load failure

manishgupta88:
Analysis: In case of data load failure on any executor, exception thrown is caught and logged as an error but the exception is not re-thrown to the spark therefore spark does not resubmits the task to the available executors again.

Solution: Catch the exception, log the error and after that re-throw the exception


gvramana:
LGTM

PR #852, opened 2016-07-22
[CARBONDATA-68] Added query statistic for detail query case measuring carbon executor time

kumarvishal09:
Added detail query statistics.


gvramana:
LGTM

PR #853, opened 2016-07-22
[CARBONDATA-95] Columns values with numeric data types are not getting parsed when included in dictionary_include

manishgupta88:
Analysis: When a numeric datatype lets say Decimal is defined for a column and the column is included as dictionary_include, then the whatever precision and scale is defined by the user is not taken into consideration and each value is accepted and dictionary is generated for that value.

Solution: Parse big decimal while generating global dictionary and dictionary look up and set the precision and scale specified by the user


core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -159,71 +155,92 @@ public static Object getDataBasedOnDataType(String data, DataType actualDataType
       return null;
     }
     try {
+      Object parsedValue = null;
```
gvramana: This function suppose to handle all types, if needed a specific function for int and long types write separate function to reflect its purpose, otherwise misleading

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -159,71 +155,92 @@ public static Object getDataBasedOnDataType(String data, DataType actualDataType
...
    *
-   * @param value    value to parse
-   * @param dataType datatype for that value
+   * @param value     value to parse
+   * @param dimension dimension to get data type and precision and scale in case of decimal
+   *                  data type
    * @return
    */
-  public static boolean validateColumnValueForItsDataType(String value, DataType dataType) {
+  public static String validateColumnValueForItsDataType(String value, CarbonDimension dimension) {
```
gvramana: change function name to normalizeColumnValueForItsDataType

processing/src/main/java/org/carbondata/processing/datatypes/PrimitiveDataType.java
```diff
@@ -151,15 +152,22 @@ public void setSurrogateIndex(int surrIndex) {
...
-  @Override
-  public void parseStringAndWriteByteArray(String tableName, String inputString,
+  @Override public void parseStringAndWriteByteArray(String tableName, String inputString,
       String[] delimiter, int delimiterIndex, DataOutputStream dataOutputStream,
       CarbonCSVBasedDimSurrogateKeyGen surrogateKeyGen) throws KettleException, IOException {
-    Integer surrogateKey = surrogateKeyGen
-        .generateSurrogateKeys(inputString, tableName + CarbonCommonConstants.UNDERSCORE + name,
-            this.getColumnId());
-    if (surrogateKey == CarbonCommonConstants.INVALID_SURROGATE_KEY) {
+    String parsedValue = DataTypeUtil.parseValue(inputString,
```
gvramana: lookup should of dimension should not be done, in map , use array based index lookup

processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1137,12 +1140,20 @@ else if(isComplexTypeColumn[j]) {
...
               surrogateKeyForHrrchy[0] =
-                  directDictionaryGenerator1.generateDirectSurrogateKey(((String) r[j]));
+                  directDictionaryGenerator1.generateDirectSurrogateKey(tuple);
               surrogateKeyGen.max[m] = Integer.MAX_VALUE;
 
             } else {
-              surrogateKeyForHrrchy[0] =
-                  surrogateKeyGen.generateSurrogateKeys(((String) r[j]), foreignKeyColumnName);
+              String parsedValue = DataTypeUtil.parseValue(tuple,
+                  data.getSurrogateKeyGen().getDimensionNameToDimensionMapping().get(
```
gvramana: Store dimension index to dimension mapping in firsttime read and use index based lookup.

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
```diff
@@ -159,71 +155,92 @@ public static Object getDataBasedOnDataType(String data, DataType actualDataType
       return null;
     }
     try {
+      Object parsedValue = null;
       switch (actualDataType) {
         case INT:
+          parsedValue = Integer.parseInt(data);
```
gvramana: Add test cases to test decimal lookup and normalization scenario.

core/src/main/java/org/carbondata/core/util/DataTypeUtil.java
manishgupta88: @gvramana ......This method is called only for parsing int and long types in dictionary exclude scenario....so I will rename this method. general purpose method is not required here.

sujith71955:
Also 2 test cases are skipped in below package, please correct it
org.carbondata.integration.spark.testsuite.complexType

manishgupta88:
@sujith71955 .....none of the test cases are failing. just before my PR one complex type bug was fixed and PRbuilder was ran. In that 2 new test cases were added. If a PR that ran just before your PR had some additional test cases than in your PR result it will display -count for the number of additional test cases that were added in the previous ran PR.

kumarvishal09:
retest this please

manishgupta88:
retest this please

PR #856, opened 2016-07-23
[BUG]Handle create table without dimensions

Jay357089:
when create table without key columns, it should failure and throw exception.

### DIFF
--- a/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
+++ b/integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
@@ -496,8 +496,15 @@ class CarbonSqlParser()
                                   tableProperties: Map[String, String]): tableModel
   = {
 
-    var (dims: Seq[Field], noDictionaryDims: Seq[String]) = extractDimColsAndNoDictionaryFields(
+    val (dims: Seq[Field], noDictionaryDims: Seq[String]) = extractDimColsAndNoDictionaryFields(
       fields, tableProperties)
+    if (dims.length == 0) {
+      throw new MalformedCarbonCommandException(s"Table ${dbName.getOrElse(
+        CarbonCommonConstants.DATABASE_DEFAULT_NAME)}.$tableName"
+        + " can not be created without key columns. Please use DICTIONARY_INCLUDE or " +
+        "DICTIONARY_EXCLUDE to set at least one key column " +
+        "if all specified columns are numeric types")
+    }
     val msrs: Seq[Field] = extractMsrColsFromFields(fields, tableProperties)
 
     // column properties
--- a/integration/spark/src/test/scala/org/carbondata/spark/testsuite/createtable/TestCreateTableSyntax.scala
+++ b/integration/spark/src/test/scala/org/carbondata/spark/testsuite/createtable/TestCreateTableSyntax.scala
@@ -99,6 +99,19 @@ class TestCreateTableSyntax extends QueryTest with BeforeAndAfterAll {
     sql("drop table if exists hivetable")
   }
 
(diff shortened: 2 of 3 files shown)


integration/spark/src/test/scala/org/carbondata/spark/testsuite/createtable/TestCreateTableSyntax.scala
```diff
@@ -99,6 +99,19 @@ class TestCreateTableSyntax extends QueryTest with BeforeAndAfterAll {
...
   }
 
+  test("create carbon table without dimensions") {
+    try {
+      sql("create table carbontable(msr1 int, msr2 double, msr3 bigint, msr4 decimal)" +
+        " stored by 'org.apache.carbondata.format'")
+      assert(false)
+    } catch {
+      case e : MalformedCarbonCommandException => {
+        println(e.getMessage)
```
Zhangshunyu: why use println

Jay357089: : ) will delete

Jay357089:
Is there anybody can review this?

Jay357089:
Is there anybody can review this?

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -496,8 +496,13 @@ class CarbonSqlParser()
...
   = {
 
-    var (dims: Seq[Field], noDictionaryDims: Seq[String]) = extractDimColsAndNoDictionaryFields(
+    val (dims: Seq[Field], noDictionaryDims: Seq[String]) = extractDimColsAndNoDictionaryFields(
       fields, tableProperties)
+    if (dims.length == 0) {
+      throw new MalformedCarbonCommandException(s"Table ${dbName.getOrElse("default")}.$tableName"
+        + " can not be created without key columns, please use DICTIONARY_INCLUDE or " +
+        "DICTIONARY_EXCLUDE to set at least one key column")
+    }
```
manishgupta88: I think We can add to the error message as....

Jay357089: Agree : ). handled.

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -496,8 +496,14 @@ class CarbonSqlParser()
                                   tableProperties: Map[String, String]): tableModel
   = {
 
-    var (dims: Seq[Field], noDictionaryDims: Seq[String]) = extractDimColsAndNoDictionaryFields(
+    val (dims: Seq[Field], noDictionaryDims: Seq[String]) = extractDimColsAndNoDictionaryFields(
       fields, tableProperties)
+    if (dims.length == 0) {
```
sujith71955: can be the first statement, since if there is no dimension defined then no need to even extract No dictionary columns, initially itself we can check and throw exception

integration/spark/src/main/scala/org/apache/spark/sql/CarbonSqlParser.scala
```diff
@@ -496,8 +496,14 @@ class CarbonSqlParser()
...
+      throw new MalformedCarbonCommandException(s"Table ${dbName.getOrElse("default")}.$tableName"
```
sujith71955: Better we can move default to a constant since in future also this constant will be used in many places because its our default table name

Jay357089:
handle comments

sujith71955:
retest this please

sujith71955:
LGTM

PR #858, opened 2016-07-23
[CARBONDATA-111]Need to catch exception in compaction  and rethrow it to spark.

ravikiran23:
Need to catch exception in compaction  and rethrow it to spark.

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
@@ -428,6 +428,7 @@ object CarbonDataRDDFactory extends Logging {
               catch {
                 case e: Exception =>
                   logger.error("Exception in compaction thread " + e.getMessage)
+                  throw e
               }
             }
             )
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
@@ -430,8 +430,8 @@ object CarbonDataRDDFactory extends Logging {
                   logger.error("Exception in compaction thread " + e.getMessage)
                   executor.shutdown()
                   // Deleting the any partially loaded data if present.
-                  // in some case the segment folder which is present in store will not have entry in
-                  // status.
+                  // in some case the segment folder which is present in store will not have entry
+                  // in status.
                   // so deleting those folders.
                   try {
                     CarbonLoaderUtil.deletePartialLoadDataIfExist(carbonLoadModel, true)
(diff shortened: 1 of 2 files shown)


ravikiran23:
issue 1 : if the Compaction Job is killed then the compaction Job is retriggering each time. 
This is wrong. if the compaction job is killed by user intentionally then need to stop the compaction .
Need to handle this case by using try and catch blocks . 

issue 2 : if the user stop the compaction Job in spark UI , but the tasks inside the job is not getting killed immediately. Need to kill the tasks immediately.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -428,6 +428,24 @@ object CarbonDataRDDFactory extends Logging {
               catch {
                 case e: Exception =>
                   logger.error("Exception in compaction thread " + e.getMessage)
+                  executor.shutdown()
```
manishgupta88: @ravikiran23 ..you can use executor.shutDownNow() here, this method will stop all the running threads submitted to the executor else job is terminated but threads submitted to it will keep running...

ravikiran23: fixed

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -428,6 +428,24 @@ object CarbonDataRDDFactory extends Logging {
...
+                  executor.shutdownNow()
+                  // Deleting the any partially loaded data if present.
+                  // in some case the segment folder which is present in store will not have entry
+                  // in status.
+                  // so deleting those folders.
+                  try {
+                    CarbonLoaderUtil.deletePartialLoadDataIfExist(carbonLoadModel, true)
+                  }
+                  catch {
+                    case e: Exception =>
```
sujith71955: Maintain a single try catch Block rather than using multiple try catches in different statement and logging same message.
Also maintain a finally block where you can call deletePartialLoadDataIfExist(), so this will make sure that the API will be called in all cases, no need to maintain one more try catch block.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -428,6 +428,24 @@ object CarbonDataRDDFactory extends Logging {
...
+                      logger
+                        .error("Exception in compaction thread while clean up of stale segments "
+                          + e
+                          .getMessage
+                        )
+                  }
+                  compactionLock.unlock()
```
sujith71955: .unlock shall be called from finally block to make sure in any kind of situation system has unlocked the resource, else there will be a scope for stale locks

sujith71955:
LGTM

PR #859, opened 2016-07-23
[ISSUE CARBONDATA-100]Implement BigInt value compression

ashokblend:
In Carbon bigint is stored as long. There is no compression done on data.
This Change is required do compression on bigint data.

### DIFF
--- a/core/src/main/java/org/carbondata/core/compression/BigIntCompressor.java
+++ b/core/src/main/java/org/carbondata/core/compression/BigIntCompressor.java
@@ -1,0 +1,213 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/carbondata/core/compression/BigIntCompressor.java
```diff
@@ -0,0 +1,213 @@
...
+          i++;
+        }
+        return intResult;
+
+      default:
+
+        long[] defaultResult = new long[value.length];
+
+        for (long a : value) {
+          defaultResult[i] = (byte) (Math.round((maxValue - a) * Math.pow(10, decimal)));
```
kumarvishal09: why this round of required as this compressor is for non decimal values

ashokblend:
closing this, as it as another pr has already been merged.

PR #860, opened 2016-07-24
[WIP]AutoDictionary

ashokblend:
If column is highcardinality and added as dictionary_include then dictionary should be generated.


kumarvishal09:
@ashokblend can u give detail description about this PR

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -319,7 +321,9 @@ class CarbonGlobalDictionaryGenerateRDD(
             // check high cardinality
             if (model.isFirstLoad && model.highCardIdentifyEnable
                 && !model.isComplexes(split.index)
-                && model.dimensions(split.index).isColumnar()) {
+                && model.dimensions(split.index).isColumnar()
+                && !model.tableProperties
+                  .isDictionaryIncludeColumn(model.dimensions(split.index).getColName)) {
```
QiangCai: suggest to define a local boolean variable outside while loop,  and use this variable in if statment
```scala
val identifyHighCard = model.isFirstLoad && model.highCardIdentifyEnable
                  && !model.isComplexes(split.index)
                 && model.dimensions(split.index).isColumnar()
                 && !model.tableProperties
                   .isDictionaryIncludeColumn(model.dimensions(split.index).getColName
```

ashokblend: @QiangCai , its handled please check.

ashokblend:
@kumarvishal09 , updated description. Please check

integration/spark/src/main/scala/org/carbondata/spark/util/GlobalDictionaryUtil.scala
```diff
@@ -270,6 +270,18 @@ object GlobalDictionaryUtil extends Logging {
...
   /**
+   *
+   * @param model
+   * @param index
+   * @return true if dimension is fit for high cardinality check
+   */
+  def isApplicableForHighCardinalityCheck(model: DictionaryLoadModel, index: Int): Boolean = {
+    model.isFirstLoad && model.highCardIdentifyEnable && !model.isComplexes(index) &&
+    model.dimensions(index).isColumnar() && !model.tableProperties
+      .isDictionaryIncludeColumn(model.dimensions(index).getColName)
```
QiangCai: In this method, I think we need to use  model.primDimensions(index) instead of  model.dimensions(index)

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
```diff
@@ -317,9 +319,7 @@ class CarbonGlobalDictionaryGenerateRDD(
             valuesBuffer ++= distinctValueList.values
             rowCount += distinctValueList.rowCount
             // check high cardinality
-            if (model.isFirstLoad && model.highCardIdentifyEnable
-                && !model.isComplexes(split.index)
-                && model.dimensions(split.index).isColumnar()) {
+            if (GlobalDictionaryUtil.isApplicableForHighCardinalityCheck(model, split.index)) {
```
QiangCai: Better to invoke this method once before line 317.

kumarvishal09:
@ashokblend can we make high cardinality detect property default value to false and when user is enabling this property then if column is in dictionary include then also we will not generate dictionary ..I think this can be a table level property

ashokblend:
@gvramana please suggest about @kumarvishal09  solution.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonGlobalDictionaryRDD.scala
ashokblend: @QiangCai it's handled please check.

ashokblend:
Retest this please

ashokblend:
Retest this please

PR #867, opened 2016-07-25
[CARBONDATA-112] Regexp_replace function is throwing NullPointerException as the expression is evaluated on executor side in carbon

nareshpr:


integration/spark/src/main/scala/org/apache/spark/sql/SparkUnknownExpression.scala
```diff
@@ -49,10 +52,13 @@ class SparkUnknownExpression(sparkExp: SparkExpression)
...
     }
     try {
-      val sparkRes = sparkExp.eval(
-        new GenericMutableRow(values.map(a => a.asInstanceOf[Any]).toArray)
-      )
-
+      val result = evaluateExpression(
+          new GenericMutableRow(values.map(a => a.asInstanceOf[Any]).toArray))
+      val sparkRes = if (isExecutor) {
+        result.asInstanceOf[InternalRow].get(0, sparkExp.dataType)
```
gvramana: test once the output of expression is null, and confirm

PR #869, opened 2016-07-25
[CARBONDATA-95] Remove map based dimension look up for getting carbon dimension

manishgupta88:
Remove map based dimension look up while getting dimension to get the datatype in surrogate key look up step as this can impact data load performance. Instead go for index based look up.


processing/src/main/java/org/carbondata/processing/surrogatekeysgenerator/csvbased/CarbonCSVBasedSeqGenStep.java
```diff
@@ -1838,34 +1836,42 @@ private void populateCarbonMeasures(String[] measures) {
...
-    for (CarbonDimension dimension : dimensionsList) {
+    CarbonDimension[] dimensionOrdinalToDimensionMapping =
+        new CarbonDimension[meta.getColumnSchemaDetailsWrapper().getColumnSchemaDetailsMap()
+            .size()];
+    List<CarbonDimension> dimListExcludingNoDictionaryColumn = dimensionsList;
+    if (null != meta.getNoDictionaryDims() && meta.getNoDictionaryDims().length() > 0) {
+      dimListExcludingNoDictionaryColumn =
+          new ArrayList<>(dimensionsList.size() - meta.noDictionaryCols.length);
+      for (CarbonDimension dimension : dimensionsList) {
+        if (!dimension.getEncoder().isEmpty()) {
```
kumarvishal09: what about complex dimension parent column??? For parent column we are not generating any dictionary file do we need to add those column??

manishgupta88: complex dimension parent column are not getting added in the array...only complex type children which have dictionary encoding are getting added to array dimensionOrdinalToDimensionMapping

kumarvishal09:
retest this please

gvramana:
Please update in previous PR that review comment is fixed and give reference.

PR #874, opened 2016-07-26
[CARBONDATA-110] delete segment selected for compaction.

ravikiran23:
if compaction is running and the user deleted the segment which is selected for compaction. then need to fail the compaction.

### DIFF
--- a/integration/spark/src/main/java/org/carbondata/spark/merger/CarbonDataMergerUtil.java
+++ b/integration/spark/src/main/java/org/carbondata/spark/merger/CarbonDataMergerUtil.java
@@ -161,6 +161,12 @@ public final class CarbonDataMergerUtil {
         for (LoadMetadataDetails loadDetail : loadDetails) {
           // check if this segment is merged.
           if (loadsToMerge.contains(loadDetail)) {
+            // if the compacted load is deleted after the start of the compaction process,
+            // then need to discard the compaction process and treat it as failed compaction.
+            if (loadDetail.getLoadStatus()
+                .equalsIgnoreCase(CarbonCommonConstants.MARKED_FOR_DELETE)) {
+              return tableStatusUpdationStatus;
+            }
             loadDetail.setLoadStatus(CarbonCommonConstants.SEGMENT_COMPACTED);
             loadDetail.setModificationOrdeletionTimesStamp(modificationOrDeletionTimeStamp);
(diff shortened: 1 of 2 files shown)


kumarvishal09:
@ravikiran23 I have one doubt segment status you are checking after compaction process is done whether any segment got deleted or not??

kumarvishal09:
@ravikiran23 Please raise JIRA issue

kumarvishal09:
@ravikiran23 Please provide more details about the solution

ravikiran23:
CARBONDATA-110 is the jira raised for this issue. 

Details are : 

if user deletes the segment already selected for compaction then compaction need to get failed.

step 1: Start the compaction which selects the segments 0, 1 .

step 2: delete the segment 1.

so that delete will immediately make the segment 1 as marked as delete. 
but the compaction is running and it will make the already  deleted segment as compacted. like undo the delete.

Behaviour should be , if any segment gets deleted after the trigger of compaction then we need to fail the compaction process.
This can be achieved by , check the segments status before updating the compaction status in table status , and if any segment is deleted then need to abort the compaction.

kumarvishal09:
retest this please

sujith71955:
Please resolve the conflict

integration/spark/src/main/java/org/carbondata/spark/merger/CarbonDataMergerUtil.java
```diff
@@ -161,6 +161,14 @@ public static boolean updateLoadMetadataWithMergeStatus(List<LoadMetadataDetails
         for (LoadMetadataDetails loadDetail : loadDetails) {
           // check if this segment is merged.
           if (loadsToMerge.contains(loadDetail)) {
+            // if the compacted load is deleted after the start of the compaction process,
+            // then need to discard the compaction process and treat it as failed compaction.
+            if (loadDetail.getLoadStatus()
+                .equalsIgnoreCase(CarbonCommonConstants.MARKED_FOR_DELETE)) {
```
sujith71955: Add proper log messages so that user can know why the compaction will be failed

sujith71955:
LGTM

integration/spark/src/main/java/org/carbondata/spark/merger/CarbonDataMergerUtil.java
ravikiran23: fixed

PR #875, opened 2016-07-27
[CARBONDATA-113] handle quering sum(decimal(30,10))+10

foryou2030:
CARBONDATA-113 url：
`https://issues.apache.org/jira/browse/CARBONDATA-113?filter=-1`

analysis:
after debuged , and found that:

1. when using decimal(27,10) (percision < 28)
logical plan keep the same with physical plan
// spark plan 
`CarbonAggregate false, [CheckOverflow((promote_precision(`cast`(SUMCarbonFinal(PartialSum#16) `as` decimal(38,10))) + 10.0000000000), DecimalType(38,10)) AS _c0#14]`

2. when using decimal(30,10)
logical plan to physical plan: the "cast ... as decimal()" has been Cut off
// spark plan 
`CarbonAggregate false, [CheckOverflow((promote_precision(SUMCarbonFinal(PartialSum#16)) + 10.0000000000), DecimalType(38,10)) AS _c0#14]`

3. for SUM + N Aggregation，the spark will cast the sum of decimal to another decimaltype
`decimal(m, n) => decimal(m+11, n)`
once the precision is bigger than 27, spark will not  cast it to decimal(m+11),
maybe because `28+11 is larger than 38(MAX_PRECISION)`

4. when the precision is bigger than 27, we should cast it to decimal(38,n) by carbon

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
@@ -116,7 +116,7 @@ case class AverageCarbon(child: Expression, castedDataType: DataType = null)
       AverageCarbonFinal(partialSum.toAttribute,
         child.dataType match {
           case IntegerType | StringType | LongType | TimestampType => DoubleType
-          case _ => child.dataType
+          case _ => CarbonScalaUtil.updateDataType(child.dataType)
         }),
       partialSum :: Nil)
   }
@@ -156,7 +156,7 @@ case class SumCarbon(child: Expression, castedDataType: DataType = null)
         if (castedDataType != null) {
           castedDataType
         } else {
-          CarbonScalaUtil.updatedDataTypeForSum(child.dataType)
+          CarbonScalaUtil.updateDataType(child.dataType)
         }),
       partialSum :: Nil)
   }
@@ -269,7 +269,7 @@ case class SumDistinctCarbon(child: Expression, castedDataType: DataType = null)
         if (castedDataType != null) {
           castedDataType
         } else {
-          CarbonScalaUtil.updatedDataTypeForSum(child.dataType)
+          CarbonScalaUtil.updateDataType(child.dataType)
         }),
       partialSum :: Nil)
   }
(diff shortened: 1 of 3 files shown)


integration/spark/src/main/scala/org/carbondata/spark/util/CarbonScalaUtil.scala
```diff
@@ -117,6 +117,18 @@ object CarbonScalaUtil {
...
   }
 
+  def updatedDataTypeForSum(
+      currentDataType: org.apache.spark.sql.types.DataType): org.apache.spark.sql.types.DataType = {
+    var newDataType: org.apache.spark.sql.types.DataType = currentDataType
+    if (currentDataType.isInstanceOf[DecimalType]) {
+      val precision = currentDataType.asInstanceOf[DecimalType].precision
+      val scale = currentDataType.asInstanceOf[DecimalType].scale
+      if (precision > 27) {
+        newDataType = DecimalType(DecimalType.MAX_PRECISION, scale)
```
manishgupta88: @foryou2030 ....
1. Please remove the hard coded check of "precision>27", this is not correct
2. We are already modifying the datatype in eval method in case class SumFunctionCarbon......check the method getDecimalDataTypeWithUpdatedPrecision...is that not solving the prupose?...is the control not going to eval method?
3. If that is not solving the purpose, then we need to handle the updating of decimal datatype for average and sum distinct scenario too.

foryou2030: 2. the control will go to eval method, and also call getDecimalDataTypeWithUpdatedPrecision method, but the  method can't resolve this issue. we need to update decimal datatype before eval.
3. i will check if average and sum distinct scenario has same problem

manishgupta88: I think then you can update method getDecimalDataTypeWithUpdatedPrecision and there you can set MAX_PRECISION directly without checking if condition and remove the code from eval method and update the casted datatype as you are doing now

foryou2030: I checked it ：
1. avg is ok, but sum distinct has same problem
2. set MAX_PRECISION directly is ok
so, i will 
1. modify sum and sum distinct by your suggestion
2. avg keep the same

foryou2030:
@manishgupta88 @QiangCai
Refreshed, pls check

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -153,7 +153,11 @@ case class SumCarbon(child: Expression, castedDataType: DataType = null)
     val partialSum = Alias(SumCarbon(child), "PartialSum")()
     SplitEvaluation(
       SumCarbonFinal(partialSum.toAttribute,
-        if (castedDataType != null) castedDataType else child.dataType),
+        if (castedDataType != null) {
+          castedDataType
+        } else {
+          CarbonScalaUtil.updatedDataTypeForSum(child.dataType)
```
manishgupta88: @foryou2030 
1. Change the updatedDataTypeForSum method name to updateDataType
2.  Handle the avg case also as you have handled for sum and sum distinct
3. Delete method getDecimalDataTypeWithUpdatedPrecision from CarbonUitl.scala class

foryou2030:
handled comments @QiangCai @manishgupta88

manishgupta88:
LGTM

QiangCai:
LGTM

PR #877, opened 2016-07-27
[bug]fix bug of duplicate rows in UnivocityCsvParser

QiangCai:
When input data file contain some characters which consist of  one more bytes, the UnivocityCsvParser will read duplicate rows from next data block.

Root cause: 
The block length in class BlockDetails is the number of bytes in the data block. But UnivocityCsvParser use a buffer stream of char, and set the limit of this char stream with the number of bytes.

Solution:
change CustomReader to byte array reader(CustomDataStream).  Then we wrap an char array reader on CustomDataStream


kumarvishal09:
@QiangCai can u add the test case for the failure scenario. I want to see whether we can fix in older code or not

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomLineReader.java
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      return true;
+    }
+  }
+
+  public Text getCurrentValue() {
+    return value;
+  }
+
+  public synchronized void close() throws IOException {
```
kumarvishal09: why we need to synchronized i cannot see any scenario where same input stream will be closed by two thread.

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomLineReader.java
```diff
@@ -0,0 +1,114 @@
...
+    if (null == this.recordDelimiterBytes) {
+      in = new LineReader(fileIn, bufferSize);
+    } else {
+      in = new LineReader(fileIn, bufferSize, this.recordDelimiterBytes);
+    }
+
+    // If this is not the first split, we always throw away first record
+    // because we always (except the last split) read one extra line in
+    // next() method.
+    if (start != 0) {
```
kumarvishal09: you should go little back, like blockOffset-1, and skip line. Otherwise end up with skipping lines.
In case of block split exactly after file deleimiter, then end up with loosing line.

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomLineReader.java
QiangCai: fixed

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomLineReader.java
QiangCai: In case of block split exactly after file deleimiter,  the last block will also read one more line from next block.

QiangCai:
retest this please

ravipesala:
@QiangCai @kumarvishal09 The way SparkCSV does is not so efficient, even I have compared the performance between SparkCSV and our code but it is not effective. Here there is duplicate string to char conversion is happening, that could be avoided by using the current code approach.
And when come to the defect, we have mistakenly taken the CustomReader with char array, but it could be simply avoided by taking `CustomDataStream extends InputStream` and wrap it with reader. I have verified by taking this approach and it works as well.

QiangCai:
ok, I change to implement CustomDataStream

QiangCai:
@ravipesala @kumarvishal09  please review.  I have simplified the modification.

QiangCai:
Modification:
1. Change CustomReader( char array reader) to CustomDataReader(byte array reader)
2. startOffset does not need to go back one byte.  we also does not need to check the end byte of block. We always read one extra line from next block.

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomDataStream.java
```diff
@@ -0,0 +1,107 @@
...
+          endOfLineFound = true;
+          this.remaining += NUMBER_OF_EXTRA_CHARACTER_TO_READ;
+        } else if (endOfLineFound) {
+          for (int i = 0; i < length; i++) {
+            if (buffer[i] == END_OF_LINE_BYTE_VALUE) {
+              this.remaining = 0;
+              return i + 1;
+            }
+          }
+          this.remaining += NUMBER_OF_EXTRA_CHARACTER_TO_READ;
```
kumarvishal09: Please correct me if I m wrong  we are reading every time 100 bytes extra then it should be this.remaining = NUMBER_OF_EXTRA_CHARACTER_TO_READ;

kumarvishal09:
@QiangCai LGTM except one comment please check

ravipesala:
@QiangCai it seems there is an issue in the code, please try with 10 million in your test code, it gives extra rows.

kumarvishal09:
I tested with 1million oscon data it is ok

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomDataStream.java
QiangCai: same, when this.remaining is 0
this.remaining += NUMBER_OF_EXTRA_CHARACTER_TO_READ;

QiangCai:
retest this please.

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomDataStream.java
kumarvishal09: If remaining is 0 remaining will be 100 but in below line it will be 200 then 300 ...so on

QiangCai:
retest this please

processing/src/main/java/org/carbondata/processing/csvreaderstep/CustomDataStream.java
kumarvishal09: Invalid comment

ravipesala:
@QiangCai Issue is with `offset` handling in `read` method of `CustomDataStream` .Please use as follows.
```
public int read(byte[] buffer, int offset, int length) throws IOException {
    if (this.remaining == 0) {
      return -1;
    } else {
      if (this.remaining < length) {
        length = (int) this.remaining;
      }

      length = this.stream.read(buffer, offset, length);
      if (length >= 0) {
        this.remaining -= length;
        if (this.remaining == 0 && !endOfLineFound) {
          if (buffer[length - 1] != END_OF_LINE_BYTE_VALUE) {
            endOfLineFound = true;
            this.remaining += NUMBER_OF_EXTRA_CHARACTER_TO_READ;
          }
        } else if (endOfLineFound) {
          for (int i = offset; i < length; i++) {
            if (buffer[i] == END_OF_LINE_BYTE_VALUE) {
              this.remaining = 0;
              return (i - offset) + 1;
            }
          }
          this.remaining += NUMBER_OF_EXTRA_CHARACTER_TO_READ;
        }
      }
      return length;
    }
  }
```

QiangCai:
int end = offset + length;
for (int i = offset; i < end; i++) {

QiangCai:
@ravipesala   10 million dataloading success, record count is right.

kumarvishal09:
@David what is the issue?

QiangCai:
FileFooter.read  throw TTransportException.

ravipesala:
@QiangCai That is the issue with local file block split. That issue has been fixed in apache github. Please use hdfs to verify the issue.

ravipesala:
LGTM.

ravipesala:
Please use 1 million in test case, otherwise it takes long time for PR builder. I will merge this PR.

QiangCai:
done

kumarvishal09:
U can set fs.local.block.size property for above issue

ravipesala:
@kumarvishal09 it is about our fact files are getting splitted when we local store path instead of hdfs, we should not do that. Any way that is a different issue and our users don't use local storage so it would not be any issue.  This issue has been handled in apache github.

ravipesala:
LGTM

PR #878, opened 2016-07-27
[CARBONDATA-76] Null values getting display in not equals filter for timestamp datatype column

manishgupta88:
Analysis: While setting surrogates for exclude filter in custom dictionary visitor, surrogate for null value is not getting added to surrogate list due to which the query displays null value.

Impact: Not equals, equals, not null and is null query on timestamp datatype column

Fix: Add surrogate value for null member in exclude filter list


sujith71955:
LGTM

PR #879, opened 2016-07-27
[CARBONDATA-114][Bug] While describing/querying the complex type dimension decimal co…

sujith71955:
Issue description:
While describing/querying the complex type dimension decimal columns system was not able to maintain the scale/precision, always default precision has been taken which needs to be handled.

Solution:
While resolving the metadata relation the decimal columns precision and scale shall be retrieved from CarbonDimension instance and has to be considered while forming the relations.


manishgupta88:
LGTM

integration/spark/src/main/scala/org/apache/spark/sql/CarbonDatasourceRelation.scala
```diff
@@ -257,6 +253,16 @@ case class CarbonRelation(
     }
   }
 
+  def addDecimalScaleAndPrecision(dimval: CarbonDimension, dataType: String): String = {
```
nareshpr: @sujith71955, Can u add testcase for describe command with complex types with decimal column as complex primitive ?

gvramana:
LGTM

PR #881, opened 2016-07-28
Fix when csv header provided in load ddl is different from schema the exception is not proper

Zhangshunyu:
1. When load data into a table, the csv header provided in LOAD DDL is not the same as table schema, it will throws exception like  "CSV File provided is not proper. Column names in schema and CSV header are not same" , which is not proper, actually it is not caused by the CSV File, it is caused by the wrong csv header provided in load ddl.
2. If both ddl and csv file not provide the file header, add the error message that "Either the ddl or the CSV file should provide CSV file header".


sujith71955:
retest this please

Zhangshunyu:
Could any one pls review this pr？

QiangCai:
LGTM

PR #882, opened 2016-07-28
CARBONDATA-117 BlockLet distribution for optimum resource usage

mohammadshahidkhan:
Problem: 
In present scenario during querying data, the data scan can be done minimum at the block level.
Suppose if the system can execute 10 parallel task and number of blocks to be scanned is 3, then only 3 cores could be utilised remaining 7 will be un-utilized.
Solution:
To optimize the resouce usage the blocks can be further divided into multiple blocks.
A block consist of multiple blocklets, these blocklet could be distributed into multiple logical blocks by assigning the startBlockletNumber and no of BlockLetsTobeScanned.


mohammadshahidkhan:
Test result for : 61700 records:
Comparison between data retrieved without blocklets distribution vs blocklets distribution
blocklet_distribution

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
```diff
@@ -69,6 +82,18 @@ public TableBlockInfo(String filePath, long blockOffset, String segmentId, Strin
     this.blockLength = blockLength;
   }
 
+  public TableBlockInfo(String filePath, long blockOffset, String segmentId, String[] locations,
```
kumarvishal09: its better to create a model object as number of parameter is increasing

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
```diff
@@ -185,6 +210,15 @@ public long getBlockLength() {
         > ((TableBlockInfo) other).blockOffset + ((TableBlockInfo) other).blockLength) {
       return 1;
     }
+    //compare the startBlockLetNumber
+    int diffStartBlockLetNumber =
```
kumarvishal09: start blocklet number and number of blocklet to scan should not participate in hashcode and equals method. please remove

core/src/main/java/org/carbondata/query/util/DataFileFooterConverter.java
```diff
@@ -95,10 +96,15 @@
         BlockIndex readBlockIndexInfo = indexReader.readBlockIndexInfo();
         blockletIndex = getBlockletIndex(readBlockIndexInfo.getBlock_index());
         dataFileFooter = new DataFileFooter();
+        TableBlockInfo tableBlockInfo = tableBlockInfoList.get(counter++);
```
kumarvishal09: This method was added to read the carbon index file main purpose of this method was to improve the driver block reading  performance basically to improve the count(*) query performance. why we are reading metadata from carbondata file It will hit block reading performance. No need to set total number of block while reading the carbonindex file as for non filter query it will be useless (we are setting -1) and for filter query can only set after applying min max in carbon query rdd

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -272,7 +270,7 @@ private void setSegmentsToAccess(Configuration configuration, List<String> segme
       }
 
       if (filterPredicates == null) {
-        return getSplitsInternal(job);
+        return getSplitsInternal2(job);
```
kumarvishal09: please give proper method name

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -442,8 +445,6 @@ private Object getFilterPredicates(Configuration configuration) {
       AbsoluteTableIdentifier absoluteTableIdentifier, FilterResolverIntf resolver,
       String segmentId) throws IndexBuilderException, IOException {
 
-    QueryStatisticsRecorder recorder = new QueryStatisticsRecorder("");
```
kumarvishal09: This code was to record driver block loading time statistic any reason why this code we have removed??

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
```diff
@@ -129,14 +129,21 @@ private void intialiseInfos() {
...
+      //if number of block is less than 0 then take end block.
+      if(numberOfBlockToScan <= 0) {
+        DataRefNode endDataBlock = finder
+            .findLastDataBlock(blockInfo.getDataBlock().getDataRefNode(), blockInfo.getEndKey());
+        numberOfBlockToScan = endDataBlock.nodeNumber() - startDataBlock.nodeNumber() + 1;
+      }
+      this.totalNumberBlockletPerSlice[index] =  numberOfBlockToScan;
       totalNumberOfNode += this.totalNumberBlockletPerSlice[index];
       blockInfo.setFirstDataBlock(startDataBlock);
-      blockInfo.setNumberOfBlockToScan(1);
```
kumarvishal09: out older approach for detail query was to scan one blocklet of each block (based on number of cores) at a time that is why we were setting 1, any reason for setting number of bocklet?? If we are setting number of blocklet then updateSliceIndexToBeExecuted this method logic will also change

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
mohammadshahidkhan: Moved blocklets related to BlockletInfos class

core/src/main/java/org/carbondata/core/carbon/datastore/block/TableBlockInfo.java
mohammadshahidkhan: I have used only startBlockletNumber. Its done intentionally to identify the unique block while dataload.
    Set<Distributable> uniqueBlocks =
        new HashSet<Distributable>(CarbonCommonConstants.DEFAULT_COLLECTION_SIZE);

core/src/main/java/org/carbondata/query/util/DataFileFooterConverter.java
mohammadshahidkhan: its added to support the blocklets distribution in non filter queries.
I could not found any other better for the same.

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputFormat.java
mohammadshahidkhan: fixed

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputFormat.java
mohammadshahidkhan: Sorry, hope this happened due to chery pic , will fix

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
mohammadshahidkhan: I have changed to 1, but yet i think we can discuss on this.

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputFormat.java
```diff
@@ -524,7 +540,7 @@ private Object getFilterPredicates(Configuration configuration) {
 
       // Add all blocks of btree into result
       DataRefNodeFinder blockFinder =
-          new BTreeDataRefNodeFinder(segmentProperties.getEachDimColumnValueSize());
+          new BTreeDataRefNodeFinder(segmentProperties.getDimensionColumnsValueSize());
```
kumarvishal09: Any reason to change the method?

core/src/main/java/org/carbondata/query/util/DataFileFooterConverter.java
kumarvishal09: For setting the number of blocklet you take the total record count of the block snd divide it by number of record per blocklet

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputFormat.java
mohammadshahidkhan: This merge problem, i think getDimensionColumnsValueSize you have change to  getEachDimColumnValueSize but this changed got reverted while chery pic will fix the same

core/src/main/java/org/carbondata/query/util/DataFileFooterConverter.java
mohammadshahidkhan: Fixed

hadoop/src/main/java/org/carbondata/hadoop/CarbonInputSplit.java
```diff
@@ -34,6 +34,10 @@
 
   private static final long serialVersionUID = 3520344046772190207L;
   private String segmentId;
+  /**
+   * Number of BlockLets in a block
+   */
+  private int numberOfBlocklets = 0;
```
kumarvishal09: do not set the default value by default value will be zero

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -79,6 +81,22 @@
 
   private static final LogService LOGGER =
       LogServiceFactory.getLogService(CarbonLoaderUtil.class.getName());
+  /**
+   * minimum no of blocklet required for distribution
+   */
+  private static int minBlockLetsReqForDistribution = 0;
```
kumarvishal09: do not set the default value by default value will be zero

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -1408,4 +1426,62 @@ public static void checkAndCreateCarbonDataLocation(String carbonStorePath, Stri
...
+   * method to distribute the blocklets of a block in multiple blocks
+   * @param blockInfoList
+   * @param defaultParallelism
+   * @return
+     */
+  public static List<Distributable> distributeBlockLets(List<TableBlockInfo> blockInfoList,
+      int defaultParallelism) {
+    LOGGER.info("No.Of Blocks before Blocklet distribution: " + blockInfoList.size());
+    List<Distributable> tableBlockInfos = new ArrayList<Distributable>();
+    if (blockInfoList.size() < defaultParallelism) {
```
kumarvishal09: One doubt if after number of blocks is One and defaultParallelism is 4 , and in one block if number of blocklet in that block is suppose 1000 in that case also we are doing blocklet distribution or not?

core/src/main/java/org/carbondata/core/carbon/datastore/block/BlockletInfos.java
```diff
@@ -0,0 +1,112 @@
...
+import java.io.Serializable;
+
+/**
+ * The class holds the blocks blocklets info
+ */
+public class BlockletInfos implements Serializable {
+  /**
+   * no of blockLets
+   */
+  private int noOfBlockLets = 0;
```
kumarvishal09: do not set the default value by default value will be zero

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -1408,4 +1426,62 @@ public static void checkAndCreateCarbonDataLocation(String carbonStorePath, Stri
...
+      int index = -1;
+      for (TableBlockInfo tableBlockInfo : blockInfoList) {
+        int noOfBlockLets = tableBlockInfo.getBlockletInfos().getNoOfBlockLets();
```
kumarvishal09: Please simplify the logic current logic is very complex. Please simplify based on out discussion

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
mohammadshahidkhan: yes we do

PR #883, opened 2016-07-28
[CARBONDATA-118] Temp location clean up in compaction

ravikiran23:
Temp locations should get deleted after the compaction is completed in RDD.

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonMergerRDD.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonMergerRDD.scala
@@ -83,12 +83,13 @@ class CarbonMergerRDD[K, V](
       }
       storeLocation = storeLocation + '/' + System.nanoTime() + '/' + theSplit.index
       CarbonProperties.getInstance().addProperty(tempLocationKey, storeLocation)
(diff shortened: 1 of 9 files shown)


integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonMergerRDD.scala
```diff
@@ -70,7 +71,21 @@ class CarbonMergerRDD[K, V](
...
+
+      val storeLocations = CarbonLoaderUtil.getConfiguredLocalDirs(SparkEnv.get.conf)
+      if (null != storeLocations && storeLocations.length > 0) {
+        storeLocation = storeLocations(Random.nextInt(storeLocations.length))
+      }
+      if (storeLocation == null) {
+        storeLocation = System.getProperty("java.io.tmpdir")
+      }
+      storeLocation = storeLocation + '/' + System.nanoTime() + '/' + theSplit.index
+      CarbonProperties.getInstance().addProperty(tempLocationKey, storeLocation)
```
gvramana: compaction should use different keys compared to dataload

ravikiran23: fixed.

gvramana:
LGTM

PR #884, opened 2016-07-28
[CARBONDATA-121]Need to identify the segments for merging each time after one compaction

ravikiran23:
Need to identify the segments for merging each time after one compaction is done.

as the user can delete some loads in between the compaction process.

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
@@ -408,7 +408,7 @@ object CarbonDataRDDFactory extends Logging {
                   .DEFAULT_COLLECTION_SIZE
               )
 
-              scanSegmentsAndSubmitJob(futureList)
+              scanSegmentsAndSubmitJob(futureList, loadsToMerge)
 
               futureList.asScala.foreach(future => {
                 future.get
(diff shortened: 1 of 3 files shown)


ravikiran23:
this includes changes for stopping the major compaction after it is successfull once. not doing major compaction continuously as it may include the new loads again for major compaction.

integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonDataRDDFactory.scala
```diff
@@ -408,16 +414,24 @@ object CarbonDataRDDFactory extends Logging {
...
+              // scan again and determine if anything is there to merge again.
               readLoadMetadataDetails(carbonLoadModel, hdfsStoreLocation)
               segList = carbonLoadModel.getLoadMetadataDetails
+              // in case of major compaction we will scan only once and come out as it will keep
+              // on doing major for the new loads also.
+              // excluding the newly added segments.
+              if (compactionModel.compactionType == CompactionType.MAJOR_COMPACTION) {
 
+                segList = CarbonDataMergerUtil
+                  .filterOutNewlyAddedSegments(segList, loadsToMerge, lastSegment)
```
gvramana: what happens if last segment is deleted

ravikiran23: it is ok because we are reading complete list of segments from the table status . even deleted segs will be read.

gvramana:
LGTM

PR #885, opened 2016-07-28
[ISSUE CARBONDATA-119]Changed locking so that zookeeper lock or hdfs lock will be available…

ashokblend:
Dictionary generation happens at executor side. Currently zookeeper is initialised at driver side and not at executor side.
We need to pass the zookeeper detail to executor to iniialise.

We need to pass hdfs temp location also to hdfs in case of hdfs lock.


gvramana:
LGTM, please rebase

ashokblend:
retest this please

QiangCai:
LGTM

PR #887, opened 2016-07-29
[CARBONDATA-122] Block distribution give second and third preferred locaiton

gvramana:
Currently as there is grouping of blocks while distribution, only first option of preferred location is given to spark. So Identify the second and third option to schedule.

### DIFF
--- a/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
+++ b/integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
@@ -264,8 +264,33 @@ class CarbonQueryRDD[V: ClassTag](
    /**
     * Get the preferred locations where to launch this task.
     */
-  override def getPreferredLocations(partition: Partition): Seq[String] = {
-    val theSplit = partition.asInstanceOf[CarbonSparkPartition]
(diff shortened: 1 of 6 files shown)


integration/spark/src/main/scala/org/carbondata/spark/rdd/CarbonQueryRDD.scala
```diff
@@ -264,8 +264,33 @@ class CarbonQueryRDD[V: ClassTag](
...
-    theSplit.locations.filter(_ != "localhost")
-  }
+   override def getPreferredLocations(partition: Partition): Seq[String] = {
+     val theSplit = partition.asInstanceOf[CarbonSparkPartition]
+     val firstOptionLocation = theSplit.locations.filter(_ != "localhost")
+     val tableBlocks = theSplit.tableBlockInfos
+     // node name and count mapping
+     val blockMap = new util.LinkedHashMap[String, Integer]()
+
+     tableBlocks.asScala.foreach(tableBlock => tableBlock.getLocations.foreach(
```
sujith71955: Grouping logic can be common between Query and load RDD

sujith71955:
retest this please

sujith71955:
LGTM

PR #888, opened 2016-07-29
[CARBONDATA-124] Exception thrown while executing drop table command in spark-sql cli

manishgupta88:
Analysis: Whenever drop table operation is performed, first the table is deleted from carbon metastore and then command is executed to delete the same table from hive metastore. While deleting from hive metastore the call comes back to carbon metastore catalog where we again check for table existence but table is not found and exception is thrown. Call comes again to carbon because the command is executed as spark sql command and not as hive sql command and we catch the logical plan when drop command is executed as spark sql command.

Impact: drop table operation

Fix: Run the drop table command as hive sql while deletiing table from hive metastore


sujith71955:
LGTM

PR #893, opened 2016-08-01
[CARBONDATA-127] Issue while type casting data read from sort temp file to big decimal type

manishgupta88:
Analysis: Whenever we perform data load operation involving huge data with decimal datatypes, then during intermediate merging pf sort temp file we are trying to typecast byte array to big decimal value after reading from object array. During this operation typecast exception is thrown.

Impact area: data load flow with huge data and measures with big decimal datatype

Fix: Typescast the object array value to byte array instaed of bigdecimal


kumarvishal09:
LGTM

PR #895, opened 2016-08-01
[Issue CARBONDATA-129]Add null check before adding to CarbonProperties.

ashokblend:
Added null check before adding to carbonproperties.


sujith71955:
LGTM

kumarvishal09:
LGTM

PR #897, opened 2016-08-01
[CARBONDATA-127] Decimal datatype result in cast exception in compaction

ravikiran23:
in case of decimal measure we will get spark Decimal data type.
but in writer it expects byte array. so converting decimal to byte array.

adding test cases for same.
and correcting equijoin test case.

### DIFF
--- a/processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
+++ b/processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
@@ -76,6 +76,8 @@ import org.carbondata.processing.store.writer.NodeHolder;
 import org.carbondata.processing.store.writer.exception.CarbonDataWriterException;
 import org.carbondata.processing.util.RemoveDictionaryUtil;
 
+import org.apache.spark.sql.types.Decimal;
+
 /**
  * Fact data handler class to handle the fact data
  */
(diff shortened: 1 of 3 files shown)


processing/src/main/java/org/carbondata/processing/store/CarbonFactDataHandlerColumnar.java
```diff
@@ -889,7 +903,15 @@ private void calculateMaxMin(Object[] max, Object[] min, int[] decimal, int[] ms
           int num = (value % 1 == 0) ? 0 : CarbonCommonConstants.CARBON_DECIMAL_POINTERS_DEFAULT;
           decimal[count] = (decimal[count] > num ? decimal[count] : num);
         } else if (type[count] == CarbonCommonConstants.BIG_DECIMAL_MEASURE) {
-          byte[] buff = (byte[]) row[count];
+          byte[] buff = null;
+          // in compaction flow the measure with decimal type will come as spark decimal.
+          // need to convert it to byte array.
+          if (this.compactionFlow) {
+            BigDecimal bigDecimal = ((Decimal) row[customMeasureIndex[i]]).toJavaBigDecimal();
```
gvramana: here it should be count , but not customMeasureIndex[i]

gvramana:
LGTM

PR #899, opened 2016-08-02
Deleted load/loads support from DDL commands

ManoharVanam:
Deleted load/loads support & using only segment/segments in all commands


integration-testcases/src/test/scala/org/carbondata/spark/testsuite/allqueries/AllDataTypesTestCase6.scala
```diff
@@ -1785,7 +1785,7 @@ class AllDataTypesTestCase6 extends QueryTest with BeforeAndAfterAll {
   test("TC_1252") {
     sql("create table table7 (imei string,AMSize string,channelsId string,ActiveCountry string, Activecity string,gamePointId decimal,deviceInformationId INT) stored by 'org.apache.carbondata.format'")
     checkAnswer(
```
sujith71955: Add testcases for identifying illegal commands which still used loads, will make our system more stable

ManoharVanam: Added test cases

sujith71955:
LGTM

PR #900, opened 2016-08-03
[CARBONDATA-138] Avg aggregation for decimal type keeping sync with hive

foryou2030:
when we do avg aggregation for decimal type:
hive wiil update the precision of the avg result. 
e.g: Avg(`decimal(30,10)`), the result will be cast to `decimal(34, 14) `
so, carbon should keep the sync with hive, also need to improve the precision

the details that copyed from spark source code:

```
case class Average(child: Expression) extends UnaryExpression with PartialAggregate1 {

  override def prettyName: String = "avg"

  override def nullable: Boolean = true

  override def dataType: DataType = child.dataType match {
    case DecimalType.Fixed(precision, scale) =>
      // Add 4 digits after decimal point, like Hive
      DecimalType.bounded(precision + 4, scale + 4)
    case _ =>
      DoubleType
  }
....
```

### DIFF
--- a/core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
+++ b/core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
@@ -25,6 +25,10 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 
+import static java.lang.Math.min;
+
+import org.apache.spark.sql.types.DecimalType;
+
 import org.carbondata.core.carbon.datastore.chunk.MeasureColumnDataChunk;
 import org.carbondata.core.constants.CarbonCommonConstants;
 import org.carbondata.core.util.DataTypeUtil;
@@ -115,7 +119,8 @@ public class AvgBigDecimalAggregator extends AbstractMeasureAggregatorBasic {
    * @return average aggregate value
    */
   @Override public BigDecimal getBigDecimalValue() {
-    int updatedScale = aggVal.scale() + 4;
+    // increase scale to avoid any precision lost in the data
+    int updatedScale = min(aggVal.scale() + 4, DecimalType.MAX_SCALE());
     return aggVal.divide(new BigDecimal(count), updatedScale, BigDecimal.ROUND_HALF_EVEN);
   }
 
@@ -140,7 +145,8 @@ public class AvgBigDecimalAggregator extends AbstractMeasureAggregatorBasic {
    * @return average value as an object
    */
   @Override public Object getValueObject() {
-    int updatedScale = aggVal.scale() + 4;
+    // increase scale to avoid any precision lost in the data
+    int updatedScale = min(aggVal.scale() + 4, DecimalType.MAX_SCALE());
     return aggVal.divide(new BigDecimal(count), updatedScale, BigDecimal.ROUND_HALF_EVEN);
   }
 
--- a/core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
+++ b/core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
@@ -115,7 +115,8 @@ public class AvgBigDecimalAggregator extends AbstractMeasureAggregatorBasic {
    * @return average aggregate value
    */
   @Override public BigDecimal getBigDecimalValue() {
-    return aggVal.divide(new BigDecimal(count), 6);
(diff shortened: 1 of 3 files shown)


foryou2030:
@QiangCai   pls  check

core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
```diff
@@ -115,7 +115,8 @@ public AvgBigDecimalAggregator() {
    * @return average aggregate value
    */
   @Override public BigDecimal getBigDecimalValue() {
-    return aggVal.divide(new BigDecimal(count), 6);
+    int updatedScale = aggVal.scale() + 4;
```
gvramana: what is this 4? Please write comment when using constant values

core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
```diff
@@ -139,7 +140,8 @@ public AvgBigDecimalAggregator() {
    * @return average value as an object
    */
   @Override public Object getValueObject() {
-    return aggVal.divide(new BigDecimal(count));
+    int updatedScale = aggVal.scale() + 4;
```
gvramana: min(precision + 4, DecimalType.MAX_PRECISION) also required here also

core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
foryou2030: ok, i 'll add comment

core/src/main/java/org/carbondata/query/aggregator/impl/AvgBigDecimalAggregator.java
foryou2030: only scale required for divide fuction. u means min(scale+ 4, DecimalType.MAX_SCALE) ? yes?

gvramana: Yes, use min function while changing scale

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -116,7 +115,14 @@ case class AverageCarbon(child: Expression, castedDataType: DataType = null)
...
         child.dataType match {
           case IntegerType | StringType | LongType | TimestampType => DoubleType
-          case _ => CarbonScalaUtil.updateDataType(child.dataType)
+          case decimal: DecimalType =>
+            val precision = decimal.asInstanceOf[DecimalType].precision
+            val scale = decimal.asInstanceOf[DecimalType].scale
+            // update precision,keep sync with hive
+            val updatedPrecision = min(precision + 4, DecimalType.MAX_PRECISION)
+            val updatedScale = min(scale + 4, DecimalType.MAX_SCALE)
+            DecimalType(updatedPrecision, updatedScale)
```
manishgupta88: @foryou2030 ...you can remove the below code
val updatedPrecision = min(precision + 4, DecimalType.MAX_PRECISION)
val updatedScale = min(scale + 4, DecimalType.MAX_SCALE)
DecimalType(updatedPrecision, updatedScale)

Instead directly write 
DecimalType.bounded(precision + 4, scale + 4)

integration/spark/src/main/scala/org/carbondata/spark/agg/CarbonAggregates.scala
```diff
@@ -116,7 +115,14 @@ case class AverageCarbon(child: Expression, castedDataType: DataType = null)
       AverageCarbonFinal(partialSum.toAttribute,
         child.dataType match {
           case IntegerType | StringType | LongType | TimestampType => DoubleType
-          case _ => CarbonScalaUtil.updateDataType(child.dataType)
+          case decimal: DecimalType =>
+            val precision = decimal.asInstanceOf[DecimalType].precision
+            val scale = decimal.asInstanceOf[DecimalType].scale
+            // update precision,keep sync with hive
```
manishgupta88: @foryou2030 ...In the comment I don't think we should mention about hive behavior, you can directly put like increase precision and scale to avoid any precision loss in the data

foryou2030: ok

foryou2030:
@gvramana pls review

PR #902, opened 2016-08-03
[CARBONDATA-135]Fixed multiple hdfs client creation issue

kumarvishal09:
Problem:When opening a input stream we are creating client every time and each time it is taking around 400 ms, In case of detail query we are opening for each blocklet and it is impacting the query performance
Solution:As query execution is a reading operation we can open only one client 
Impact Area:Query execution and compaction

### DIFF
--- a/core/src/main/java/org/carbondata/core/datastorage/store/impl/FileFactory.java
+++ b/core/src/main/java/org/carbondata/core/datastorage/store/impl/FileFactory.java
@@ -124,7 +124,7 @@ public final class FileFactory {
       case HDFS:
       case VIEWFS:
         Path pt = new Path(path);
-        FileSystem fs = pt.getFileSystem(configuration);
+        FileSystem fs = FileSystem.get(configuration);
         FSDataInputStream stream = fs.open(pt);
         return new DataInputStream(new BufferedInputStream(stream));
       default:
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/carbondata/core/datastorage/store/impl/FileFactory.java
```diff
@@ -166,7 +166,7 @@ public static DataInputStream getDataInputStream(String path, FileType fileType,
       case HDFS:
       case VIEWFS:
         Path pt = new Path(path);
-        FileSystem fs = pt.getFileSystem(configuration);
+        FileSystem fs = FileSystem.get(configuration);
```
manishgupta88: @kumarvishal09 ...same thing needs to be handled in other methods of FileFactory (like getDataOutputStream) and in AbstractDFSCarbonFile constructor where for each new file object we are creating the new path and getting filesystem from that path

kumarvishal09: I have not tested the impact with output stream that's why i have not added

gvramana: @manishgupta88 This can be handled in separate PR.

gvramana:
This PR is handling only for query flow scan part.
There are other read flows like dictionary reading flow, CarbonHdfsFile cases can handle in another PR.

PR #903, opened 2016-08-03
[CARBONDATA-136]Fixed Query data mismatch issue after compaction

kumarvishal09:
Problem: During compaction we are calling query execution, for compaction we are doing merge sort for merge sort all the bucket must be sorted in so as we are calling query execution we can only execute one blocklet at a time, currently we are executing multiple blocklet of difference blocks. Because of this bucket records are not sorted.
Solution:Need to execute one blocklet a time during compaction

### DIFF
--- a/core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
+++ b/core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
@@ -97,10 +97,18 @@ public abstract class AbstractDetailQueryResultIterator extends CarbonIterator {
   public AbstractDetailQueryResultIterator(List<BlockExecutionInfo> infos,
       QueryExecutorProperties executerProperties, QueryModel queryModel,
       InternalQueryExecutor queryExecutor) {
+    // below code will be used to update the number of cores based on number
+    // records we
+    // can keep in memory while executing the query execution
     int recordSize = 0;
-    String defaultInMemoryRecordsSize =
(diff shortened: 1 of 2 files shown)


sraghunandan:
LGTM

core/src/main/java/org/carbondata/query/carbon/result/iterator/AbstractDetailQueryResultIterator.java
```diff
@@ -97,15 +97,25 @@
...
+    } else {
+      String defaultInMemoryRecordsSize =
+          CarbonProperties.getInstance().getProperty(CarbonCommonConstants.INMEMORY_REOCRD_SIZE);
+      if (null != defaultInMemoryRecordsSize) {
+        try {
+          recordSize = Integer.parseInt(defaultInMemoryRecordsSize);
+        } catch (NumberFormatException ne) {
+          LOGGER.error("Invalid inmemory records size. Using default value");
+          recordSize = CarbonCommonConstants.INMEMORY_REOCRD_SIZE_DEFAULT;
+        }
```
manishgupta88: @kumarvishal09 ...In case when inmemory record size is not set in the query model neither it is configured then in that case record size will be 0. So either the null check can be removed or default property can be passed while getting the configured value from carbonproperties instance

kumarvishal09: ok

kumarvishal09: This property can not be null any time And if it is null then numberOfCores will be one but it is better to handle this

gvramana:
LGTM

PR #904, opened 2016-08-03
[CARBONDATA-137]Fixed detail limit query statistics issue

kumarvishal09:
Problem:In case of detail query with limit it total query execution time in statistics is printing negative values as wrong method is getting called
Solution:Need to call correct method


gvramana:
LGTM

PR #905, opened 2016-08-04
[CARBONDATA-139]Inconsistency during sortindex file reading

ashokblend:
Earlier we used to check size of dictionary and then create sortIndex file,to read, by appending size of dictionary e.g columid_offset.sortindex

During sort index reading phase, read offset from dictionary chunk meta and then get respective file.


manishgupta88:
LGTM

PR #906, opened 2016-08-04
[Bug] Fix the bug of Block Prune

Zhangshunyu:
## Why raise this pr:
During block prune, endkey is always only decided by the last filter expression, this is a bug
and can lead wrong result,
For example, when load data whose dimension column is 120000 lines of 'a', 120000 lines of 'b', 120000 lines of 'c', if query like "select * from tablename where colname='c' or colname='b' or colname='a'" only 120000lines 'a' will be selected because of wrong endkey.

## How to solve this:
Fix the end key consider all the filter expression end key get max and start key get (min - 1) for each column level, using this to produce a new start key an a new endkey.

For more details please look at the test case.

### DIFF
--- a/core/src/main/java/org/carbondata/query/filters/measurefilter/util/FilterUtil.java
+++ b/core/src/main/java/org/carbondata/query/filters/measurefilter/util/FilterUtil.java
@@ -90,7 +90,7 @@ import org.carbondata.query.filter.resolver.RowLevelFilterResolverImpl;
 import org.carbondata.query.filter.resolver.resolverinfo.DimColumnResolvedFilterInfo;
 import org.carbondata.query.schema.metadata.DimColumnFilterInfo;
 
-public final class FilterUtil {
+public final class  FilterUtil {
   private static final LogService LOGGER =
       LogServiceFactory.getLogService(FilterUtil.class.getName());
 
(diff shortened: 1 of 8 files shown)


Zhangshunyu:
retest this please

core/src/main/java/org/carbondata/query/filters/measurefilter/util/FilterUtil.java
```diff
@@ -693,27 +693,14 @@ public static DimColumnFilterInfo getFilterListForAllMembersRS(Expression expres
...
-      if (isExcludePresent) {
-        continue;
-      }
-      getStartKeyBasedOnFilterResoverInfo(dimensionFilter, startKey);
+  public static void getStartKey(Map<CarbonDimension, List<DimColumnFilterInfo>> dimensionFilter,
+      AbsoluteTableIdentifier tableIdentifier, long[] startKey, SegmentProperties segmentProperties,
+      List<long[]> startKeyList) throws QueryExecutionException {
+    for(int i = 0; i < startKey.length; i++) {
+      // The min surrogate key is 2, set it as the init value for starkey of each column level
+      startKey[i] = 2;
```
eason-lyx: NULL value will be 1

Zhangshunyu:
retest this please

Zhangshunyu:
This pr has been merged into apache github，so i l close here.

PR #909, opened 2016-08-08
[CARBONDATA-152]Double min max difference compression issue.

ashokblend:
1.While compressing maxdifference typecast with proper DataType
2.While writting we follow like
   a. diff with max value , say d = max - x
   b. multiply with 10^decimal i.e w = d*10^decimal
3.Similarly while reading we should follow like
   a. divide written value 10^decimal i.e d = w/10^decimal
   b. diff with max i.e x = max -d
4. There is problem in java with double subtraction
   i.e double value = 3.141818 - 0.000610
      java will result 3.1412080000000002
    To avoid this we are converting to BigDecimal and then doing subtraction
5.Changed logic to calculate no of integer after decimal.

### DIFF
--- a/core/src/main/java/org/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
+++ b/core/src/main/java/org/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
@@ -19,6 +19,8 @@
 
 package org.carbondata.core.datastorage.store.compression.type;
 
+import java.math.BigDecimal;
+
 import org.carbondata.common.logging.LogService;
 import org.carbondata.common.logging.LogServiceFactory;
 import org.carbondata.core.datastorage.store.compression.Compressor;
@@ -93,7 +95,9 @@ public class UnCompressNonDecimalMaxMinByte
       if (value[i] == 0) {
         vals[i] = maxValue;
       } else {
-        vals[i] = (maxValue - value[i]) / Math.pow(10, decimalVal);
+        BigDecimal diff = new BigDecimal(Double.toString(value[i] / Math.pow(10, decimalVal)));
+        BigDecimal max = new BigDecimal(Double.toString(maxValue));
+        vals[i] = max.subtract(diff).doubleValue();
       }
 
     }
--- a/core/src/main/java/org/carbondata/core/util/ValueCompressionUtil.java
+++ b/core/src/main/java/org/carbondata/core/util/ValueCompressionUtil.java
@@ -19,6 +19,7 @@
 
 package org.carbondata.core.util;
 
+import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
@@ -482,13 +483,16 @@ public final class ValueCompressionUtil {
   private static Object compressNonDecimalMaxMin(DataType changedDataType, double[] value,
       int decimal, double maxValue) {
     int i = 0;
+    BigDecimal max = new BigDecimal(Double.toString(maxValue));
     switch (changedDataType) {
       case DATA_BYTE:
 
         byte[] result = new byte[value.length];
 
         for (double a : value) {
-          result[i] = (byte) (Math.round((maxValue - a) * Math.pow(10, decimal)));
+          BigDecimal val = new BigDecimal(Double.toString(a));
+          double diff = max.subtract(val).doubleValue();
+          result[i] = (byte) (Math.round(diff * Math.pow(10, decimal)));
           i++;
         }
         return result;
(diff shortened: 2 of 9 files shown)


Zhangshunyu:
Could you please provide a test case? Especially the values MAX - MIN has little delta.

core/src/main/java/org/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
```diff
@@ -93,7 +93,7 @@
       if (value[i] == 0) {
         vals[i] = maxValue;
       } else {
-        vals[i] = (maxValue - value[i]) / Math.pow(10, decimalVal);
```
eason-lyx: double datatype calculate(like sub) will lost precision.
like 3.1415-0.01 = 3.1413999999998
suggest use bigDecimal to handle the calculate,then convert to double.

ashokblend:
Retest this please

core/src/main/java/org/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
ashokblend: @eason-lyx  this compression is for double data type only. If difference of double falls in range of byte then only this piece of code will get executed.

ashokblend:
@Zhangshunyu  added test case

core/src/main/java/org/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
eason-lyx: yes,i know.
u can try double value:
3.1411
3.1412
3.1413
3.1414
3.1415
then,after query it may shows
3.1411
3.1412
3.14129999998
3.1414
3.14149999998

so better to make it as string ,and use decimal compute the sub,then toDoubleType.

eason-lyx: also value is byte value,i think we can no directly use ,need to change to double value.

ashokblend: @eason-lyx I will try with this example.

integration/spark/src/test/scala/org/carbondata/spark/testsuite/detailquery/ValueCompressionDataTypeTestCase.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+class ValueCompressionDataTypeTestCase extends QueryTest with BeforeAndAfterAll {
+  val tempDirPath = "./src/test/resources/temp"
+
+  override def beforeAll {
+    FileFactory.mkdirs(tempDirPath,FileType.LOCAL)
+  }
+
+  test("ActualDataType:double,ChangedDatatype:Short,CompressionType:NonDecimalMaxMin") {
+    val tempFilePath = "./src/test/resources/temp/dsnondecimalmaxmin.csv"
```
mohammadshahidkhan: @ashokblend  Please add commit the dsnondecimalmaxmin.csv file

ashokblend: @mohammadshahidkhan this file is generated at runtime and also gets deleted at runtime.

core/src/main/java/org/carbondata/core/datastorage/store/compression/type/UnCompressNonDecimalMaxMinByte.java
ashokblend: @eason-lyx please check now.

ashokblend:
retest this please

eason-lyx:
LGTM

core/src/main/java/org/carbondata/core/util/ValueCompressionUtil.java
```diff
@@ -482,13 +483,16 @@ private static Object compressNonDecimal(DataType changedDataType, double[] valu
...
     switch (changedDataType) {
       case DATA_BYTE:
 
         byte[] result = new byte[value.length];
 
         for (double a : value) {
-          result[i] = (byte) (Math.round((maxValue - a) * Math.pow(10, decimal)));
+          BigDecimal val = new BigDecimal(Double.toString(a));
+          double diff = max.subtract(val).doubleValue();
+          result[i] = (byte) (Math.round(diff * Math.pow(10, decimal)));
```
ashokblend: @gvramana @kumarvishal09 
I feel Math.round will always lead to precision error. Any idea why we are rounding it ?

ashokblend:
closing this issue as it is merged in incubator

PR #910, opened 2016-08-08
Remove failed test case which need to be adapt to new thrift format in the future.

Zhangshunyu:
The test cases need to be adapt to new thrift format in the future.


kumarvishal09:
@Zhangshunyu Better delete test file and data file

Zhangshunyu:
@kumarvishal09 Done, pls check.

PR #912, opened 2016-08-09
Fix the failed test case using new thrift format to generate new  .carbondata file

Zhangshunyu:
Fix the failed test case using new thrift format to generate new .carbondata file instead of the older one.

PR #913, opened 2016-08-09
CARBONDATA-153 Record count is not matching while loading the data when one data node went down in HA setup

mohammadshahidkhan:
Record count is not matching while loading the data when one data node went down in HA setup

As per previous implementation : 
Scenario: 
No. of Running executors =  3
No. Data Nodes                =  2
Total Unique blocks          = 96
Then as per the previous implementation no. of blocks per node = 32.
While assigning blocks only 66 blocks getting allocated amongs two execuotors. 
Third executor is not getting any blocks since we are considering only node local allocation.
Solution:
// so now we have a map of node vs blocks. allocate the block as per the order
    createOutputMap(nodeBlocksMap, blocksPerNode, uniqueBlocks, nodeAndBlockMapping, activeNodes);

    // if the nodeBlocksMap size is than the activeNodes then populate the remaining active nodes
    // in nodeBlockMap
    assignLeftOverNodes(nodeBlocksMap, activeNodes);

After doing node block mapping if the the size of nodeBlocks is less than the activeNodes list size then we will map the remaining activeNodes in the  nodeBlocksMap, after  that the  assignLeftOverBlocks 
will take care the assignment of remaining blocks.
// if any blocks remain then assign them to nodes in round robin.
    assignLeftOverBlocks(nodeBlocksMap, uniqueBlocks, blocksPerNode);

### DIFF
--- a/integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
+++ b/integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
@@ -1022,8 +1022,11 @@ public final class CarbonLoaderUtil {
     Map<String, List<Distributable>> mapOfNodes =
         CarbonLoaderUtil.nodeBlockMapping(blockInfos, noOfNodesInput, activeNode);
     int taskPerNode = parallelism / mapOfNodes.size();
-    //assigning non zero value to noOfTasksPerNode
-    int noOfTasksPerNode = taskPerNode == 0 ? 1 : taskPerNode;
+    // assigning non zero value to noOfTasksPerNode
+    // Multiplying the taskPerNode to devide the long running job into small job so that the
+    // job's of the node running for long time  can be re allocated to the node having no jobs.
+    // (for example non-nodelocal jobs takes more time than the nodelocal jobs)
+    int noOfTasksPerNode = taskPerNode == 0 ? 2 : 2*taskPerNode;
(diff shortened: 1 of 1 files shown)


mohammadshahidkhan:
Failling Test cases will be fixed by : #912

mohammadshahidkhan:
@gvramana Please review this PR

integration/spark/src/main/java/org/carbondata/spark/load/CarbonLoaderUtil.java
```diff
@@ -1119,13 +1119,36 @@ public static Dictionary getDictionary(CarbonTableIdentifier tableIdentifier,
...
 
   /**
+   * the method populates the left over active nodes in node block map.
+   *
+   * @param nodeBlocksMap
+   * @param activeNodes
+   */
+  private static void assignLeftOverNodes(Map<String, List<Distributable>> nodeBlocksMap,
+      List<String> activeNodes) {
+    if (null != activeNodes && activeNodes.size() > nodeBlocksMap.size()) {
```
gvramana: Merge this logic with previous function, so that we don't need to loop two times

mohammadshahidkhan: Fixed review comments

mohammadshahidkhan:
closing as raised the similar in incubator
https://github.com/apache/incubator-carbondata/pull/77

